Issue #3, opened 2021-11-14
Could not retrieve watch page info: Invalid value "undefined" for header "x-goog-visitor-id"


UnbreakCode:
Downloading from YouTube is not possible due to this error.

"Could not retrieve watch page info: Invalid value "undefined" for header "x-goog-visitor-id" "

LuanRT:
Yea, I've noticed this too but it would usually happen randomly. It seems like sometimes YouTube doesn't provide the visitorData and some other fields in the Innertube configuration json. And that ends up causing a lot of problems— I will try to find a more efficient way to fetch that data.

LuanRT:
Closing this for now as it is no longer a problem when it comes to downloading, however I should probably investigate it more since YouTube might complain about it not being provided.

PR #4, opened 2021-11-14
fixed x-goog-visitor-id for downloader

UnbreakCode:
fixed this issue https://github.com/LuanRT/YouTube.js/issues/3


LuanRT:
Seems good, merging.
And thanks!

PR #8, opened 2021-12-13
Include available video qualities to metadata

stranothus:
The playerResponse streamingData adaptiveFormats are filtered to include only those which
include a qualityLabel. This array is then mapped to an array of qualityLabels and sorted
from lowest to highest quality. A Set is created then spread syntax is used to remove duplicates.

### DIFF
--- a/lib/Constants.js
+++ b/lib/Constants.js
@@ -141,7 +141,7 @@ module.exports = {
       metadata.publish_date = data.microformat.playerMicroformatRenderer.publishDate || 'N/A';
       metadata.upload_date = data.microformat.playerMicroformatRenderer.uploadDate || 'N/A';
       metadata.keywords = data.videoDetails.keywords || [];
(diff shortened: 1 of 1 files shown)


lib/Constants.js
```diff
@@ -141,6 +141,7 @@ module.exports = {
       metadata.publish_date = data.microformat.playerMicroformatRenderer.publishDate || 'N/A';
       metadata.upload_date = data.microformat.playerMicroformatRenderer.uploadDate || 'N/A';
       metadata.keywords = data.videoDetails.keywords || [];
+      metadata.available_qualities = [...new Set(data[2].playerResponse.streamingData.adaptiveFormats.filter(v => v.qualityLabel).map(v => v.qualityLabel).sort((a, b) => +a.replace(/\D/gi, "") - +b.replace(/\D/gi, "")))]
```
LuanRT, line 144: Amazing contribution!
Just one thing; the Innertube API sends a slightly different response for the desktop version of YouTube— so change this line right here to ``data.streamingData.adaptiveFormats`` and it should be good.
Thanks!

LuanRT (changes requested):

lib/Constants.js
stranothus: Thank you, the change has been made.

LuanRT (approved):

PR #61, opened 2022-05-30
typings: improve type definitions

iShibi:
## Description

Make method parameter types more specific for better TypeScript support.

## Checklist:


LuanRT:
This is nice, but it would eventually get overwritten as type definitions are generated from jsdoc.

Issue #45, opened 2022-05-01
Support 'highest' quality


trullock:
## Expected Behavior

To be able to request the `highest` quality available and have it automatically selected

## Current Behavior

Currently you have to mimic the inner workings of `#chooseFormat` in order to replicate this logic and then manually iterate down the potential qualities to see if it exists. This should be encapsulated within the library.

## Sample Code

it would be nice to be able to do this:

```
{
    quality: 'highest'
}
```

trullock:
*activity*

Wykerd:
I'm closing this issue since the above PR implements this suggestion and has been available in the main branch for quite a while now.

PR #76, opened 2022-06-16
feat: download video directly from VideoInfo

Wykerd:
As suggested in #45, this also implements a new "best" and
"bestefficiency" format selector.

@LuanRT please check that the new chooseFormat and download methods look ok. I've tested it on 2 videos and seems to work fine.

### DIFF
--- a/lib/Innertube.js
+++ b/lib/Innertube.js
@@ -29,6 +29,7 @@ const Channel = require('./parser/youtube/Channel');
 const Playlist = require('./parser/youtube/Playlist');
 const FilterableFeed = require('./core/FilterableFeed');
 const TabbedFeed = require('./core/TabbedFeed');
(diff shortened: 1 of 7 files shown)


Wykerd:
I have no idea why the download test still fails, its downloading perfectly now.

LuanRT (approved):

LuanRT (approved):

LuanRT:
Works like a charm!
But I think we can make it even faster — you see, for downloads the initial request to the endpoint `/player` is already enough, so we can safely omit the request (`/next`) used for additional data. 

I already Implemented the `getBasicInfo()` method and will commit the code in a bit.

LuanRT:
Ok, after some more tests I think we're done here. Merging!

PR #79, opened 2022-06-20
feat(player): improved decipher logic

Wykerd:
- Improve the deciphering logic for Signatures and NTokens.
- This makes NToken transforms more than 20x faster (on my M1 Pro its down from 1.7ms to 0.06ms)
- It also improves caching of the player drastically, by only keeping the processed responses in binary format, bringing down the cache per player from 1.8MB to less than 400 bytes

These improvements where originally written in TypeScript but I did not think it would be appropriate to add a build step to this project, so I compiled it down to JavaScript before adding to this project. [You can view the original files here](https://gist.github.com/Wykerd/6a97a810340f5382d9c38360bd12182c)


LuanRT:
Great! Do note however that Google is constantly tweaking the microlanguage in the NToken algorithm, I see this adds a bit of complexity around that and might make it more difficult to fix something if they push a major change. Also, maybe we should include the original typescript files in the same alongside the compiled deciphers?

LuanRT (approved):

LuanRT (approved):

PR #91, opened 2022-07-05
Make project multiplatform

MasterOfBob777:
## Description

This disconnects the project from node, and makes the project multi platform. This adds esbuild as a build system and separates the test system into platform specific pieces. This also adds a few dependencies so that might be something you want to avoid. 

## Type of change
- [X] New feature (non-breaking change which adds functionality)
- [?] This change requires a documentation update (it now works on browser, so potentially)

## Checklist:

- [X] My code follows the style guidelines of this project
- [X] I have performed a self-review of my own code
- [X] I have commented my code, particularly in hard-to-understand areas
- [?] I have made corresponding changes to the documentation (see above)
- [X] My changes generate no new warnings
- [X] I have added tests that prove my fix is effective or that my feature works
- [X] New and existing unit tests pass locally with my changes
- [X] I have checked my code and corrected any misspellings


LuanRT:
This looks great! I'll review the changes and do some testing. Meanwhile, could you fix the conflicts? 

Thank you.

MasterOfBob777:
So, unfortunately bundling is a tough problem, and we can't really use `--keep-names` because that just assures that the legacy variable `name` is the same throughout minification which doesn't really help us in this case. The best solution I could come up with is to include source maps which will map the minified code to the actual source, and in environments like the browser it will automatically look the same.
Types should work after this change, unless you use explicit imports of `build/node.js`, `build/browser.js`, etc. which sadly for us because our tests do not use a bundler, means that types will be unavailable there, but to any end user it should look normal. Just an aside, the issue with `idb` was just a missing `lib.dom.d.ts` because typescript thought it was just a node environment.

LuanRT (approved):

PR #93, opened 2022-07-08
Convert the project into typescript

MasterOfBob777:
## Description

This attempted to convert the entire project to typescript. This is a big change but I think it will improve the code quality of the library. It also is an essential part of my goal to convert this project to [AssemblyScript](https://www.assemblyscript.org/) so that I can use the (currently extremely alpha) [Rosetta](https://github.com/NativPlayer/rosettac) compiler to convert this into Kotlin so that there are bindings for android. That change itself is most likely out of scope of this library, although it might be beneficial to the library's performance, regardless that change won't be complete for a good amount of time because there are a lot of huge issues converting a dynamically typed javascript-y library into a strictly typed language with a similar but not identical standard library. Back on topic, this is just part 1 of that change, and simply intends to convert the project into typescript.

This currently is not finished in the slightest.

## Type of change

- [X] New feature (non-breaking change which adds functionality)
- [X] This change requires a documentation update

## Checklist:

- [ ] My code follows the style guidelines of this project
- [ ] I have performed a self-review of my own code
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [ ] My changes generate no new warnings
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] New and existing unit tests pass locally with my changes
- [ ] I have checked my code and corrected any misspellings


lib/core/Actions.ts
```diff
@@ -566,15 +607,16 @@ class Actions {
...
+   * 
    */
-  async getSearchSuggestions(client, query) {
+  async getSearchSuggestions(client: string, query: string): Promise<APIResponse<object>> {
     if (![ 'YOUTUBE', 'YTMUSIC' ].includes(client))
-      throw new Utils.InnertubeError('Invalid client', client);
+      throw new InnertubeError('Invalid client', client);
 
     const response = await ({
+      // NOTE: this is an instance of Request, so something is wrong
```
LuanRT, line 619: A `Request` instance is just an Axios instance, hence why one can make request like that.

lib/core/Actions.ts
```diff
@@ -644,15 +685,18 @@ class Actions {
...
-    return [
-      'FElibrary',
-      'FEhistory',
-      'FEsubscriptions',
-      'SPaccount_notifications',
-      'SPaccount_privacy',
-      'SPtime_watched'
-    ].includes(id);
+  #needsLogin(id: string) {
+    switch (id) {
```
LuanRT, line 689: I believe includes was more suitable here, a whole switch may be too verbose. What do you think?

lib/Innertube.ts
```diff
@@ -131,12 +208,11 @@ class Innertube {
...
   /**
    * Signs out of an account.
-   *
-   * @returns {Promise.<{ success: boolean, status_code: number }>}
    */
-  async signOut() {
+  async signOut(): Promise<{ success: boolean, status_code: number }> {
     if (!this.logged_in) throw new InnertubeError('You are not signed in');
 
+    // NOTE: this isn't a method? is this supposed to be OAuth#revokeCredentials?
```
LuanRT, line 215: See #90, I probably forgot to rename that.

lib/core/Actions.ts
LuanRT: > The main reason was performance, although this might be an unnecessary optimization. Because with the switch it is (depending on the engine), probably just a lookup table, versus a throwaway allocation and loop over the elements of the array, on top of the `Object.is` equality check. 

I get that but it wouldn't make a lot of difference as far as I'm concerned. In this particular situation, a switch statement is probably overkill. Especially when all we're doing is checking whether an ordinary string exists in an array, which is what `includes()` was made for.

LuanRT: > Also this does bring up the point, should I continue to make small changes like this to the code? This is just mainly moving types from here to there, but I think it would be beneficial to do stuff like this as I'm forced to look over every file anyways.

Sure.

MasterOfBob777: Okay, I'll revert the change.

lib/core/OAuth.ts
```diff
@@ -165,7 +167,7 @@ class OAuth {
     this.#credentials = {
       access_token: response.data.access_token,
       refresh_token: response.data.refresh_token || this.credentials.refresh_token,
-      expires: expiration_date
+      expires_in: expiration_date
```
LuanRT, line 170: This is intentionally named `expires`. The expiration time is converted from seconds to a date, so it wouldn't make sense to have the original name there.

MasterOfBob777: Ohh okay, I was going to leave a comment but it seemed like a simple mistake.

MasterOfBob777:
> This will probably be helpful to you: https://gist.github.com/Wykerd/6a97a810340f5382d9c38360bd12182c
> 
> It's the original TypeScript files I wrote for the NToken and Signature deciphering classes.

Sweet! Thanks!

PR #97, opened 2022-07-11
Better cross runtime support

Wykerd:
## Description

This builds on efforts of #91 and intends to make the library work on browsers and Deno (possibly CFWorkers) by replacing as much dependencies as possible with equivalent Web APIs available on all platforms.

This is a major change which will break event and stream APIs as these are to be replaced by their Web API equivalents.

## Type of change

## Changes:
- [x] Remove node APIs
  - [x] crypto -> SubtleCrypto
  - [x] stream -> Web Stream API
  - [x] events -> EventTarget
  - [x] make sure buffer is polyfilled on all platforms
- [x] Replace Axios with Fetch API - Fetch is available on more platforms than XMLHTTPRequest
- [x] Documentation for Deno usage
- [x] Update tests for browsers
- [x] Documentation for browser usage
- [ ] Tests for Deno? 

EDIT: will not support Bun.js at this time, too experimental at this moment.


Wykerd (commented):
I now realize that this is going to cause massive merge conflicts with #93, but how bad could it be 😅

LuanRT:
> events -> EventTarget

Maybe we could work on a custom implementation of EventEmitter? It seems quite simple and would probably be better compared to the verbosity & weirdness of the EventTarget API.

LuanRT:
> I now realize that this is going to cause massive merge conflicts with #93, but how bad could it be 😅

Haha there's already a quite big conflict due to 68cb841c0058ec5aac7b79b7736fdf57b825da3c (which was an inevitable change as we had lots of dead code laying around). I guess we could help but unfortunately he's working on a branch outside of the project.

Wykerd (commented):
How about this for EventEmitter? Best of both worlds haha

LuanRT (commented):
Awesome!

lib/parser/youtube/VideoInfo.ts
```diff
@@ -0,0 +1,414 @@
...
+            this.player_overlays = Parser.cast_response(next.player_overlays, PlayerOverlay);
+            this.basic_info.like_count = Parser.cast_response(this.primary_info?.menu?.top_level_buttons?.get({ icon_type: 'LIKE' }), ToggleButton, false)?.like_count;
+            this.basic_info.is_liked = Parser.cast_response(this.primary_info?.menu?.top_level_buttons?.get({ icon_type: 'LIKE' }), ToggleButton, false)?.is_toggled;
+            this.basic_info.is_disliked = Parser.cast_response(this.primary_info?.menu?.top_level_buttons?.get({ icon_type: 'DISLIKE' }), ToggleButton, false)?.is_toggled;
+            const comments_entry_point = Parser.cast_response(results.get({ target_id: 'comments-entry-point' }), ItemSection, false);
+            this.comments_entry_point_header = Parser.cast_response(comments_entry_point?.contents?.get({ type: 'CommentsEntryPointHeader' }), CommentsEntryPointHeader, false);
+            /**
+             * @type {import('../classes/LiveChat')}
+             */
+            // TODO: this makes no sense, LiveChat is not in the parsers list
```
LuanRT, line 109: I'm a bit confused — what happened here?

Wykerd: There's no LiveChat parser in classes, the memo can thus never contain it. There's a LiveChat class in lib/parser/youtube/LiveChat though but not sure how you want to instantiate it.

LuanRT: > There's no LiveChat parser in classes, the memo can thus never contain it. 

[There actually is](https://github.com/LuanRT/YouTube.js/blob/main/lib/parser/classes/LiveChat.js), maybe it got accidentally deleted?

Wykerd: Oh must have missed it when I wrote that, I'll fix that in next push!

Wykerd:
I'm going quite out of the scope of this PR by doing so much TS conversion too, but I cannot write my new changes in TS without then also converting some of the other parts of the project to TS too.

LuanRT:
> I'm going quite out of the scope of this PR by doing so much TS conversion too, but I cannot write my new changes in TS without then also converting some of the other parts of the project to TS too.

At this point this probably supersedes @MasterOfBob777's PR.

Wykerd:
> > I'm going quite out of the scope of this PR by doing so much TS conversion too, but I cannot write my new changes in TS without then also converting some of the other parts of the project to TS too.
> 
> At this point this probably supersedes @MasterOfBob777's PR.

There's no way I'm rewriting all the parsers to TS in this PR though, so there's still a lot of work to be done if we want the entire codebase to be TS, for now I've enabled allowJS in the tsconfig

MasterOfBob777:
Wow! This is very cool! As @Wykerd mentioned I do think my PR has its place, but I think it is probably best to wait until this PR is completed to avoid merge conflict hell.
I also just want to mention it might be better to compile for each platform separately with esbuild constants, instead of programmatically figuring out the runtime, as it increases bundle size by quite a bit if you include unnecessary code in their respective platforms (Also the `exports` field in `package.json` can also achieve a similar effect by swapping out modules by path.) Regardless this isn't a huge issue at the end of the day, because it would barely effect performance, as V8 and other engines should just realize those branches are impossible to reach and "remove" the check, and basically achieve the same thing. 
Those are just my two cents though it might not be the right decision in this case.

lib/parser/youtube/VideoInfo.ts
```diff
@@ -1,6 +1,6 @@
 import { InnertubeError, ObservedArray } from "../../utils/Utils";
 import Parser, { YTNode } from "../index.js";
-import LiveChat from "./LiveChat.js";
+import LiveChat from "../classes/LiveChat";
```
LuanRT, line 3: Note that the `getLiveChat()` method needs the other LiveChat class too.

lib/parser/classes/NavigationEndpoint.ts
```diff
@@ -2,16 +2,41 @@ import Parser, { ParsedResponse } from "../index.js";
...
 class NavigationEndpoint extends YTNode {
+    payload;
+    dialog;
+    metadata: {
+        url?: string;
+        api_url?: string;
+        page_type?: string;
+        send_post?: boolean; // TODO: is this boolean?
+    };
+    // TODO: these should be given proper types, currently infered
```
LuanRT, line 17: Pretty sure we will be getting rid of those soon. The “`callTest()`” experiment is much more flexible than implementing every single endpoint manually (which would eventually end up creating one big chungus class). So no need to worry about giving them proper types IMHO.

LuanRT, line 17: Pretty sure we will be getting rid of those soon. The “`callTest()`” experiment is much more flexible than implementing every single endpoint manually (which would eventually end up creating one big chungus class). So no need to worry about giving them proper types IMHO.

LuanRT, line 17: Note: My GitHub was acting up so had to delete and add this comment again. Anywho, original comment:

Pretty sure we will be getting rid of those soon. The “callTest()” experiment is much more flexible than implementing every single endpoint manually (which would eventually end up creating one big chungus class). So no need to worry about giving them proper types IMHO.

lib/parser/classes/MusicResponsiveListItem.ts
```diff
@@ -0,0 +1,175 @@
+// TODO: this needs a refactor
+// seems like a mess to use
```
LuanRT, line 2: YouTube Music uses 1 single renderer for all kinds of items, and it's a bit inconsistent. Agreed, nonetheless.

Wykerd:
Got basic browser example up and running 🎉

Wykerd:
image

Wykerd:
image

LuanRT:
Wow! Astounding work here! It's only been 4 days and this PR is almost complete, really impressive @Wykerd 😆

Wykerd:
Node 12 does not have EventTarget, should we support it? Guess I could polyfill it?

LuanRT:
> Node 12 does not have EventTarget, should we support it? Guess I could polyfill it?

Not sure if we should. Up until now only one person asked for Node12 support, so it might not be worth it at all.

Wykerd:
Theres a bunch more APIs only available from Node 16 so I'll need to do some additional work in order to support them. Undici for example only implements fetch for node 16 and later.

Wykerd:
The current LTS version of Node is v16, so might be able to get away with making it a requirement, otherwise I'll need to  write some fallback logic for older node versions.

LuanRT:
Ah that's a bummer, I think the minimum should be v14. If that's too much work then we can make v16 a requirement for now.

Wykerd:
I can polyfill `EventTarget` on older versions, and swap out undici for node-fetch but then the video downloading methods no longer work since node-fetch is not spec compliant and returns its streams as node`Readable` streams instead of web `ReadableStream` 

I could offer two versions of the `VideoInfo#download` method, something like `VideoInfo#downloadNode` for use when you need to support versions of node earlier than v16?

LuanRT:
Nope, don't waste your time on that. Let's stick with Node 16 for the moment.

Wykerd:
> Nope, don't waste your time on that. Let's stick with Node 16 for the moment.

I've added a way to provide a custom fetch implementation which should be a decent compromise. This way users may bring their own fetch on older node versions, or for proxying request.

Wykerd:
> See the old codebase, that was used to debug InnerTube requests, it would print every request's url and body if set to true.

Yes, but it doesn't do anything anymore since I rewrote the requests to use fetch. I could add it back?

LuanRT:
Did a little testing with OAuth and LiveChat but none work correctly. It seems that EventTarget's `CustomEvent` is not available in Node yet. 

See https://github.com/nodejs/node/issues/40678

Wykerd:
Should probably do some more tests to make sure I'm not missing anything, but this should be ready to merge now.

LuanRT:
This seems to return undefined in Node, which in turn throws an error when using `UniversalCache`:
https://github.com/LuanRT/YouTube.js/blob/79c6ca883554863220d026050a860ba7bbb99240/lib/utils/Cache.ts#L34-L35

lib/core/PlaylistManager.ts
```diff
@@ -93,16 +73,19 @@ class PlaylistManager {
 
     const payload = {
       playlistId: playlist_id,
-      actions: []
+      actions: [] as {
+                action: string;
+                setVideoId: string;
+            }[]
```
MasterOfBob777, line 79: Perhaps extract the type to make this a tiny bit more clear?

lib/core/PlaylistManager.ts
```diff
@@ -142,17 +125,22 @@ class PlaylistManager {
 
     const payload = {
       playlistId: playlist_id,
-      actions: []
+      actions: [] as {
+                action: string,
+                setVideoId?: string,
+                movedSetVideoIdPredecessor?: string
+            }[]
```
MasterOfBob777, line 132: Same as above, extract type?

Wykerd:
Whoops that broke deno and browser again 🙃

lib/parser/README.md
```diff
@@ -30,6 +45,183 @@ Unlike `parse`, this can be used to parse the entire response object.
...
+
+// Sometimes we can expect multiple types of nodes
+const results = node.asOneOf<TwoColumnSearchResults | VideoList>([TwoColumnSearchResults, VideoList]);
+
+// similarly, we can check if the node is of a certain type
+if (node.isOneOf<TwoColumnSearchResults | VideoList>([TwoColumnSearchResults, VideoList])) {
+  // Again no casting is needed, the node is already of the correct type
+  const results = node;
+}
+```
```
MasterOfBob777, line 117: Just noticed this in the documentation, but the type parameters here could be inferred by typescript, if the signature was slightly modified to use varargs instead. Just a thought, mostly syntactic sugar.
```ts
function isOneOf<T extends YTNode, K extends YTNodeConstructor<T>[]>(...types: K): K[number] {
  // ...
}
```

lib/utils/Constants.ts
```diff
@@ -0,0 +1,122 @@
+
```
MasterOfBob777, line 1: For each of these perhaps an `as const` annotation for each exported.

lib/utils/HTTPClient.ts
```diff
@@ -0,0 +1,144 @@
...
+      credentials: 'include',
+      redirect: input instanceof Request ? input.redirect : init?.redirect || 'follow',
+      // Copy the request properties from the input
+      integrity: input instanceof Request ? input.integrity : init?.integrity,
+      method: input instanceof Request ? input.method : init?.method,
+      mode: input instanceof Request ? input.mode : init?.mode,
+      referrer: input instanceof Request ? input.referrer : init?.referrer,
+      referrerPolicy: input instanceof Request ? input.referrerPolicy : init?.referrerPolicy,
+      signal: input instanceof Request ? input.signal : init?.signal,
+      keepalive: input instanceof Request ? input.keepalive : init?.keepalive
```
MasterOfBob777, line 116: This check should be extracted into a variable.

LuanRT:
I think `Innertube#session.signIn()` doesn't doesn't wait for authentication to finish.

Ex:
```js
//...

// Triggers the auth event and asks the user to authorize the authentication.
await yt.session.signIn(undefined); 

// This will return true even though authorization hasn't been granted yet.
// In fact, this should not be printed at all since the auth flow hasn't finished and any InnerTube call will inevitably fail.
console.log(yt.session.logged_in);
```

lib/utils/Utils.ts
```diff
@@ -0,0 +1,343 @@
...
+    '70', '71', '72', '73', '74', '75', '76', '77', '78', '79', '7a', '7b', '7c', '7d', '7e', '7f',
+    '80', '81', '82', '83', '84', '85', '86', '87', '88', '89', '8a', '8b', '8c', '8d', '8e', '8f',
+    '90', '91', '92', '93', '94', '95', '96', '97', '98', '99', '9a', '9b', '9c', '9d', '9e', '9f',
+    'a0', 'a1', 'a2', 'a3', 'a4', 'a5', 'a6', 'a7', 'a8', 'a9', 'aa', 'ab', 'ac', 'ad', 'ae', 'af',
+    'b0', 'b1', 'b2', 'b3', 'b4', 'b5', 'b6', 'b7', 'b8', 'b9', 'ba', 'bb', 'bc', 'bd', 'be', 'bf',
+    'c0', 'c1', 'c2', 'c3', 'c4', 'c5', 'c6', 'c7', 'c8', 'c9', 'ca', 'cb', 'cc', 'cd', 'ce', 'cf',
+    'd0', 'd1', 'd2', 'd3', 'd4', 'd5', 'd6', 'd7', 'd8', 'd9', 'da', 'db', 'dc', 'dd', 'de', 'df',
+    'e0', 'e1', 'e2', 'e3', 'e4', 'e5', 'e6', 'e7', 'e8', 'e9', 'ea', 'eb', 'ec', 'ed', 'ee', 'ef',
+    'f0', 'f1', 'f2', 'f3', 'f4', 'f5', 'f6', 'f7', 'f8', 'f9', 'fa', 'fb', 'fc', 'fd', 'fe', 'ff'
+  ];
```
MasterOfBob777, line 140: Just because this is a large lookup table maybe move this to constants?

Wykerd:
> I think `Innertube#session.signIn()` doesn't doesn't wait for authentication to finish.
> 
> Ex:
> 
> ```js
> //...
> 
> // Triggers the auth event and asks the user to authorize the authentication.
> await yt.session.signIn(undefined); 
> 
> // This will return true even though authorization hasn't been granted yet.
> // In fact, this should not be printed at all since the auth flow hasn't finished and any InnerTube call will inevitably fail.
> console.log(yt.session.logged_in);
> ```

It does not, it only waits until the first auth event is fired with the login code

Wykerd:
Should I make it wait instead? Probably makes more sense if it does.

lib/utils/Utils.ts
```diff
@@ -0,0 +1,343 @@
...
+/**
+ *
+ */
+export function getRuntime(): Runtime {
+  if ((typeof process !== 'undefined') && (process?.versions?.node))
+    return 'node';
+  if (Reflect.has(globalThis, 'Deno'))
+    return 'deno';
+  return 'browser';
+}
```
MasterOfBob777, line 279: Maybe make this check compile time?

lib/utils/Utils.ts
```diff
@@ -0,0 +1,343 @@
...
+
+/**
+ *
+ */
+export function isServer() {
+  return [ 'node', 'deno' ].includes(getRuntime());
+}
+
```
MasterOfBob777, line 287: Above.

LuanRT:
> Should I make it wait instead? Probably makes more sense if it does.

Yup, not doing so would probably cause some confusion.

lib/utils/Utils.ts
```diff
@@ -0,0 +1,343 @@
...
+            `${arr_headers.map(([ key, value ]) => `    ${key}: ${value}`).join('\n')}\n` :
+            '    (none)\n'
+        }  body:\n${
+          init?.body ? `${typeof init.body === 'string' ?
+            headers.get('content-type') === 'application/json' ?
+              JSON.stringify(JSON.parse(init.body), null, 2) :
+              init.body :
+            '    <binary>'
+          }` : '    (none)'}`
+  );
```
MasterOfBob777, line 340: This is fine but it might be better to move the logic out of this and have the print statement be easy to read.

MasterOfBob777 (commented):
Just a few small comments I had. Nothing that seem like a huge issue, or a glaring bug, but since this a large PR I most definitely missed something, and frankly I don't have a large amount of experience with the library.

lib/core/PlaylistManager.ts
Wykerd: These types aren't used anywhere else, only in that one function, so probably unnecessary to extract.

lib/parser/README.md
MasterOfBob777: Ah so maybe then wrap it in a `Return<T>`? I'd have to play around with it but I think maybe that would work?

lib/utils/Utils.ts
MasterOfBob777: My thoughts are mainly to improve code cleanliness, but honestly you are right.

lib/utils/Utils.ts
MasterOfBob777: I think most bundling tools nowadays support `package.exports` and as long as that works, well, it is just as simple as importing it.

lib/core/PlaylistManager.ts
MasterOfBob777: Alright! Seems good

lib/Innertube.ts
```diff
@@ -0,0 +1,223 @@
...
+    this.interact = new InteractionManager(this.session.actions);
+    this.music = new YTMusic(this.session);
+    this.actions = this.session.actions;
+  }
+  static async create(config: InnertubeConfig = {}) {
+    return new Innertube(await Session.create(config));
+  }
+  /**
+   * Retrieves video info.
+   * @param video_id
```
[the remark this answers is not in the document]
MasterOfBob777: [This](https://www.npmjs.com/package/eslint-plugin-tsdoc) looks like a good project.

lib/utils/Utils.ts
Wykerd: I still feel it might be beneficial to not bundle everything and let the user handle that. (For example, if someone wants to use just the parsers and not the rest of the modules in a project)

lib/parser/README.md
Wykerd: Would be great if you could figure out how to do this. Think we can keep it like this for this PR and fix at a later date then?

MasterOfBob777: Yeah I think that's appropriate.

lib/utils/Utils.ts
MasterOfBob777: Okay, fair point.

Wykerd:
> Just a few small comments I had. Nothing that seem like a huge issue, or a glaring bug, but since this a large PR I most definitely missed something, and frankly I don't have a large amount of experience with the library.

I do think there will be some small bugs that slipped through the cracks here and there, especially in the JS code. They should be easy to find once the project is completely migrated to TS.

lib/parser/classes/NavigationEndpoint.ts
Wykerd: 👍

Wykerd:
Whoops assigned this to myself accidentally 🤣

lib/Innertube.ts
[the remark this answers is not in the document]
Wykerd: I've replaced the plugin 👍

Wykerd:
I think this PR is ready to be merged now. Mind giving it a look @LuanRT ?

index.ts
```diff
@@ -2,7 +2,7 @@ import { getRuntime } from './lib/utils/Utils';
 
 // Polyfill fetch for node
 if (getRuntime() === 'node') {
-  const undici = require('undici');
+  const undici = Reflect.get(module, 'require')('undici');
```
LuanRT, line 5: This does not work when you try to require/import YouTube.js externally, it throws `Error: Cannot find module 'undici'` even though undici is in fact present in `node_modules`. Changing it back to `require('undici')` works just fine.

Wykerd: Reverted 👍

LuanRT:
I don't see any other major bug, so merging now. Thanks @Wykerd!

Issue #94, opened 2022-07-09
Upload youtube video


naptestdev:
### Describe your suggestion

According to Youtube Data API v3
<img width="841" alt="image" src="https://user-images.githubusercontent.com/88878684/178101160-b0b54219-28ef-40d1-b7bd-851f3e859de9.png">

So it should be great if you can use to internal Innertube API to allow uploading videos without the restriction.

Thanks a lot.

### Checklist

- [X] I am running the latest version.
- [X] I checked the documentation and found no answer.
- [X] I have searched the existing issues and made sure this is not a duplicate.
- [X] I have provided sufficient information.

LuanRT:
JFYI I found out the upload endpoints used by YouTube Studio require the browser to solve a botguard challenge,  which is out of scope for this project. On the bright side of things, however, the Android app ones seem to be free of any hurdles so we'll probably use that.

LuanRT:
A little update:
See #115

PR #115, opened 2022-07-24
feat: add support for uploading videos

LuanRT:
## Description

This adds support for uploading videos.

Related: #94

## Type of change

## Checklist:

- [ ] I have performed a self-review of my own code
- [ ] I have made corresponding changes to the documentation

### DIFF
--- a/src/core/Studio.ts
+++ b/src/core/Studio.ts
@@ -52,7 +52,7 @@ class Studio {
 
     return response;
   }
-  
+
   /**
    * Uploads a video to YouTube.
    * @example
@@ -67,9 +67,9 @@ class Studio {
 
     if (upload_result.status !== 'STATUS_SUCCESS')
       throw new InnertubeError('Could not process video.');
-      
+
     const response = await this.#setVideoMetadata(initial_data, upload_result, metadata);
-    
+
     return response;
   }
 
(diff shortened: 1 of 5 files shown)


src/core/Studio.ts
```diff
@@ -31,6 +52,114 @@ class Studio {
 
     return response;
   }
+
+  /**
+   * Uploads a video to YouTube.
+   * @example
+   * ```ts
+   * const buffer = fs.readFileSync('./my_awesome_video.mp4');
+   * const response = await yt.studio.upload(buffer, { title: 'Wow!' });
```
Wykerd, line 61: Should this not be 
```ts
const response = await yt.studio.upload(new Uint8Array(buffer.buffer), { title: 'Wow!' });
```
since the type of the parameter is Uint8Array and not a Node.js Buffer?

LuanRT: > Should this not be
> 
> ```ts
> const response = await yt.studio.upload(new Uint8Array(buffer.buffer), { title: 'Wow!' });
> ```
> 
> since the type of the parameter is Uint8Array and not a Node.js Buffer?

Doesn't `Buffer` extend JavaScript's `Uint8Array`? From my understanding TypeScript doesn't, and shouldn't complain about it.

PR #121, opened 2022-07-27
feat: allow users to cache OAuth credentials

Wykerd:
Use `UniversalCache` instance to cache user credentials

Opt-in via `OAuth#cacheCredentials()`. `OAuth#init()` will check the cache before starting the OAuth flow.

### DIFF
--- a/src/core/OAuth.ts
+++ b/src/core/OAuth.ts
@@ -43,9 +43,44 @@ class OAuth {
    */
   async init(credentials?: Credentials) {
     this.#credentials = credentials;
-    if (!credentials) {
-      await this.#getUserCode();
+    if (this.validateCredentials()) {
+      this.#session.emit('auth', {
+        credentials: this.#credentials,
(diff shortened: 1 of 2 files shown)


MasterOfBob777:
Isn't it a bit unsafe storing credentials on disk in plaintext? I'm not quick sure what the alternative would be but it raises red flags in my brain.

Wykerd:
> Isn't it a bit unsafe storing credentials on disk in plaintext? I'm not quite sure what the alternative would be but it raises red flags in my brain.

This is not much different to using something like a .env file to store your credentials, so I don't see this as too big an issue. It is an opt-in feature since you need to explicitly call `OAuth#cacheCredentials` so users who feel its too insecure won't be affected by this change.

LuanRT:
Lgtm! This may come in handy when experimenting with the library.

LuanRT (approved):

PR #135, opened 2022-08-11
Fix build error caused by `music#Playlist.getRelated()`

patrickkfkan:


patrickkfkan:
Unfortunately, checks did not pass because of the ntoken issue. I suppose we should wait until that issue is resolved.

PR #150, opened 2022-08-23
Fix token refining. Adding replace where breaking on JSON.parse

seomikewaltman:
NToken.js@#getTrasnformationData was throwing errors when trying to download videos. I traced it to Utils.js@refineNTokenData and added a replace() to get the string to JSON.parse correctly 👍🏼 

Fixes # (issue)

## Type of change


LuanRT:
Wait, is this still a problem? I'm pretty confident we disabled nsig deciphering a few days ago due to changes by YouTube. 

If that's the case, you can try reinstalling v2 to stay up to date with the latest commits.

chadacious:
So is there a working version of ntoken and sig at this point?

seomikewaltman:
I was using the tagged 1.4.3. My bad! I'll switch to using the latest 👍🏼

LuanRT:
> So is there a working version of ntoken and sig at this point?

Yes, latest v2 and 1.4.4 on NPM.

PR #151, opened 2022-08-24
feat: add `MusicSortFilterButton`

patrickkfkan:
A `MusicSortFilterButton` pulls up a menu much like a dropdown. Example of one is the country selector found on YouTube Music -> Explore -> Charts. The following code illustrates usage of `MusicSortFilterButton`:

```
// Endpoint for YouTube Music Charts page
const endpoint = new NavigationEndpoint({
  browseEndpoint: {
    browseId: 'FEmusic_charts'
  }
});

const fetchAndPrint = async (endpoint) => {
  console.log('Fetching Charts...')
  const response = await endpoint.call(youtube.actions, 'YTMUSIC');
  const page = Parser.parseResponse(response.data);
  const sections = page.contents_memo.getType(SectionList)?.[0]?.contents?.array() || [];
  console.log('Got the following sections:');
  console.log(JSON.stringify(sections));
  return page;
}

const page = await fetchAndPrint(endpoint);

// The Charts page shows a country selector, which is a MusicSortFilterButton.
// For convenience sake, we will just get it from the page's `contents_memo`.

const countrySelector = page.contents_memo.getType(MusicSortFilterButton)?.[0];

if (countrySelector) {
  // Get the currently selected option

  const selectedIndex = countrySelector.menu.options.findIndex((item) => item.selected);
  const selectedCountry = countrySelector.menu.options[selectedIndex];
  console.log('Currently selected country is: ' + selectedCountry.title);

  // Randomly select an option. Note the conditional check for item type, since an item 
  // in `MusicSortFilterButton.menu.options` can either be a `MusicMultiSelectMenuItem`
  // or `MusicMenuItemDivider`. `MusicMultiSelectMenuItem` is the one we are interested
  // in, since it contains the endpoint for getting updated content.

  const randomCountry = countrySelector.menu.options.find(
    (item, index) => item.type === 'MusicMultiSelectMenuItem' &&
      Math.random() > 0.5 && index !== selectedIndex); // Hey, what are the chances of random() always < 0.5 :)

  if (randomCountry) {
    console.log('Let\'s change country to: ' + randomCountry.title);
    await fetchAndPrint(randomCountry.endpoint);  // Charts page with randomCountry selected
  }
}
```

As far as coding is concerned, implementation of `MusicSortFilterButton` requires mutations data from `response.data`, which is not available within the parser. This is therefore done in the `Parser` class after `MusicSortFilterButton` has been constructed. See `Parser#applyMutations()`. I do not know if this deviates from the coding guidelines of the library, but I cannot think of a more appropriate way to do it...

### DIFF
--- a/src/core/Actions.ts
+++ b/src/core/Actions.ts
@@ -6,9 +6,10 @@ import Parser, { ParsedResponse } from '../parser/index';
 import { hasKeys, InnertubeError, MissingParamError, uuidv4 } from '../utils/Utils';
 
 export interface BrowseArgs {
-  params?: string;
+  params?: string | null;
   is_ytm?: boolean;
   is_ctoken?: boolean;
+  form_data?: {};
   client?: string;
 }
 
@@ -92,6 +93,10 @@ class Actions {
       data.continuation = id;
     } else {
       data.browseId = id;
+    }
+
+    if (args.form_data) {
+      data.formData = args.form_data;
     }
 
     if (args.client) {
--- a/src/parser/classes/MusicDetailHeader.ts
+++ b/src/parser/classes/MusicDetailHeader.ts
@@ -19,7 +19,7 @@ class MusicDetailHeader extends YTNode {
   badges;
   author?: {
     name: string;
-    channel_id: string;
+    channel_id: string | undefined;
     endpoint: NavigationEndpoint | undefined;
   };
   menu;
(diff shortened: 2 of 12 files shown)


LuanRT:
Nice! There will be some changes in `NavigationEndpoint` soon so I think we'll have to do something about that `Parser#applyMutations()` method later. But for now it looks good. Merging.

PR #152, opened 2022-08-24
Fix `music#library.getArtists()` and add `MusicShelf.bottom_button`

### LINKED ISSUE #143: (V2) Wrong IDs in `music#Library.getArtists()`
### Steps to reproduce

```
const library = youtube.music.getLibrary();
const artists = await library.getArtists();
const artist = artists.items[0];
console.log(artist);

// Result:
{
  item_type: artist
  id: ....
  name: ...
  ...
}

```

### Failure Logs

```shell
N/A
```

### Expected behavior

`artist.id` should be the ID of the artist's channel

### Current behavior

`artist.id` is NOT the ID of the artist's channel

### Version

Edge

### Anything else?

As it turns out, `artist.id` is actually the ID of a browse endpoint that points to the list of the artist's songs. To get the actual channel ID of the artist, it would be necessary to parse from this endpoint.

`music#Library.getArtists()` would therefore have to be updated to reflect this. I would be against the idea of fetching the channel ID by parsing said endpoint at the time of `getArtists()`, because that would increase fetch time and also run the risk of 429 server responses due to sending too many requests within a short period. 

I shall see what can be done about this. You can assign the issue to me.

### Checklist

patrickkfkan:
Kind of fixes #143 by returning items with `item_type` of  'library_artist'  instead of 'artist'. Items of this type have an `endpoint` that leads to a page listing of the artist's songs (equivalent to clicking an artist in Library -> Artists). I think, at this stage, it would have to be up to the user to call the endpoint to obtain further content, including the artist's channel_id. I am however open to ideas on how this part can be better presented to the user.

PR also adds `bottom_button` to `MusicShelf`. If you follow the endpoint of a library artist, you will see one such button with the 'See All By Artist' text.

### DIFF
--- a/src/parser/classes/MusicResponsiveListItem.ts
+++ b/src/parser/classes/MusicResponsiveListItem.ts
@@ -86,10 +86,13 @@ class MusicResponsiveListItem extends YTNode {
         this.#parsePlaylist();
         break;
       case 'MUSIC_PAGE_TYPE_ARTIST':
-      case 'MUSIC_PAGE_TYPE_LIBRARY_ARTIST':
       case 'MUSIC_PAGE_TYPE_USER_CHANNEL':
         this.item_type = 'artist';
         this.#parseArtist();
+        break;
+      case 'MUSIC_PAGE_TYPE_LIBRARY_ARTIST':
+        this.item_type = 'library_artist';
+        this.#parseLibraryArtist();
         break;
       default:
         if (this.#flex_columns[1]) {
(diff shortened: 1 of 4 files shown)


LuanRT:
> I think, at this stage, it would have to be up to the user to call the endpoint to obtain further content, including the artist's channel_id.

That's right, we shouldn't overcomplicate things so let's just leave that to the user.

> PR also adds bottom_button to MusicShelf. If you follow the endpoint of a library artist, you will see one such button with the 'See All By Artist' text.

Great! Looks good but there seems to be a small conflict in that particular file.

LuanRT (approved):

patrickkfkan:
Should be resolved now. Thanks.

Issue #186, opened 2022-09-14
Live Chat - Implement moderation 


LuanRT:
### Describe your suggestion

Right now we only have `<livechat>#sendMessage(text)`, things like pinning messages, moderation, etc haven't been implemented yet. Adding these would simplify the creation of applications such as moderation bots.

### Checklist

- [X] I am running the latest version.
- [X] I checked the documentation and found no answer.
- [X] I have searched the existing issues and made sure this is not a duplicate.
- [X] I have provided sufficient information.

PR #202, opened 2022-09-23
Live Chat - Implement moderation

akkadaska:
## Description

I implemented moderation of live chat, which is mentioned at issue #186.

- add `LiveChat.getItemMenu(item: ChatItemHasMenuEndpoint)`
  You can call this method with `ChatItemHasMenuEndpoint`: chat item which has `menu_endpoint` member such as 
  `LiveChatTextMessage`, and get all endpoints for moderating the chat item whose type is `Promise<Menu>`. You can call the 
moderation endpoint in the `Menu` instanse you got by following `LiveChat.selectButtonItem()` method.

- add `LiveChat.selectButtonItem(button: Button)`
  You can call chat moderation endpoint whose type is `Button` (or its sub class such as `MenuServiceItem`) with this method.
  The modelation endpoint comes from the response of  `LiveChat.getItemMenu()`, `LiveChatTextMessage.inline_action_buttons` or `LiveChatAutoModMessage.moderation_buttons`.

- fix some bugs

Here is a code example of new methods: Remove all LiveChatTextMessage that you have permittion for remove.
```typescript
liveChat.on('chat-update', async (chatAction: ChatAction) => {
  // only LiveChatTextMessage is filtered and assigned to liveChatTextMessage
  const menu = await liveChat.getItemMenu(liveChatTextMessage);
  const removeEndpoint = menu.items.filter((item) => {
      if (!item.is(MenuServiceItem)) {
        return false;
      }
      return item.as(MenuServiceItem).icon_type === 'DELETE';
    })[0];
  if (removeEndpoint) {
    liveChat.selectButtonItem(removeEndPoint.as(MenuServiceItem));
  }
}
```

Implement #186 

## Type of change

Please delete options that are not relevant.

## Checklist:

### DIFF
--- a/src/parser/classes/NavigationEndpoint.ts
+++ b/src/parser/classes/NavigationEndpoint.ts
@@ -230,12 +230,6 @@ class NavigationEndpoint extends YTNode {
         params: data.sendLiveChatVoteEndpoint.params
       };
     }
-
-    if (data?.liveChatItemContextMenuEndpoint) {
-      this.live_chat_item_context_menu = {
-        params: data.liveChatItemContextMenuEndpoint.params
-      };
-    }
   }
 
   /**
@@ -249,6 +243,8 @@ class NavigationEndpoint extends YTNode {
         return '/player';
       case 'watchPlaylistEndpoint':
         return '/next';
+      case 'liveChatItemContextMenuEndpoint':
+        return 'live_chat/get_item_context_menu';
     }
   }
 
@@ -297,6 +293,15 @@ class NavigationEndpoint extends YTNode {
       const response = await actions.engage(this.metadata.api_url, { video_id: this.like.target.video_id, params: this.like.params });
       return response;
     }
+
+    if (this.live_chat_item_context_menu) {
+      if (!this.metadata.api_url)
+        throw new Error('Live Chat Item Context Menu endpoint requires an api_url, but was not parsed from the response.');
+      const response = await actions.livechat(this.metadata.api_url, {
+        params: this.live_chat_item_context_menu.params
+      });
+      return response;
+    }
   }
 
   async call(actions: Actions, client: string | undefined, parse: true) : Promise<ParsedResponse | undefined>;
@@ -307,7 +312,7 @@ class NavigationEndpoint extends YTNode {
     if (parse && result)
       return Parser.parseResponse(result.data);
 
-    return this.#call(actions, client);
+    return result;
   }
 }
 
--- a/src/parser/classes/livechat/items/LiveChatPaidSticker.ts
+++ b/src/parser/classes/livechat/items/LiveChatPaidSticker.ts
@@ -23,6 +23,7 @@ class LiveChatPaidSticker extends YTNode {
   sticker: Thumbnail[];
   purchase_amount: string;
   context_menu: NavigationEndpoint;
+  menu_endpoint?: NavigationEndpoint;
   timestamp: number;
 
   constructor(data: any) {
@@ -42,7 +43,8 @@ class LiveChatPaidSticker extends YTNode {
     this.author_name_text_color = data.authorNameTextColor;
     this.sticker = Thumbnail.fromResponse(data.sticker);
     this.purchase_amount = new Text(data.purchaseAmountText).toString();
-    this.context_menu = new NavigationEndpoint(data.contextMenuEndpoint);
+    this.menu_endpoint = new NavigationEndpoint(data.contextMenuEndpoint);
+    this.context_menu = this.menu_endpoint;
     this.timestamp = Math.floor(parseInt(data.timestampUsec) / 1000);
   }
 }
--- a/src/parser/youtube/ItemMenu.ts
+++ b/src/parser/youtube/ItemMenu.ts
@@ -5,11 +5,12 @@ import Button from '../classes/Button';
 import Menu from '../classes/menus/Menu';
 import MenuServiceItem from '../classes/menus/MenuServiceItem';
 import NavigationEndpoint from '../classes/NavigationEndpoint';
+import { ObservedArray, YTNode } from '../helpers';
 
 class ItemMenu {
   #page: ParsedResponse;
   #actions: Actions;
-  #menu: Menu;
+  #items: ObservedArray<YTNode>;
   constructor(data: ParsedResponse, actions: Actions) {
     this.#page = data;
     this.#actions = actions;
@@ -17,7 +18,7 @@ class ItemMenu {
     if (!menu || !menu.is(Menu)) {
       throw new InnertubeError('Response did not have an "live_chat_item_context_menu_supported_renderers" property. The call may have failed.');
     }
-    this.#menu = menu.as(Menu);
+    this.#items = menu.as(Menu).items;
   }
 
   async selectItem(icon_type: string): Promise<ParsedResponse>
@@ -27,7 +28,7 @@ class ItemMenu {
     if (item instanceof Button) {
       endpoint = item.endpoint;
     } else {
-      const button = this.#menu.items.find((button) => {
+      const button = this.#items.find((button) => {
         if (!button.is(MenuServiceItem)) {
           return false;
         }
@@ -39,16 +40,13 @@ class ItemMenu {
       }
       endpoint = button.as(MenuServiceItem).endpoint;
     }
-    if (!endpoint.metadata.api_url) {
-      throw new InnertubeError('Response did not have an "api_url" property. The call may have failed.');
-    }
-    const response = await this.#actions.execute(endpoint.metadata.api_url, {...endpoint.payload, parse: true});
+    const response = await endpoint.callTest(this.#actions, { parse: true });
 
     return response;
   }
 
-  menu(): Menu {
-    return this.#menu;
+  items(): ObservedArray<YTNode> {
+    return this.#items;
   }
 
   page(): ParsedResponse {
--- a/src/parser/youtube/ItemMenu.ts
+++ b/src/parser/youtube/ItemMenu.ts
@@ -1,23 +1,28 @@
-import { ParsedResponse } from '..';
 import Actions from '../../core/Actions';
-import { InnertubeError } from '../../utils/Utils';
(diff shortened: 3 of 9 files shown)


src/parser/classes/NavigationEndpoint.ts
```diff
@@ -230,12 +230,6 @@ class NavigationEndpoint extends YTNode {
         params: data.sendLiveChatVoteEndpoint.params
       };
     }
-
-    if (data?.liveChatItemContextMenuEndpoint) {
```
akkadaska, line 234: This was duplicated.

src/parser/classes/NavigationEndpoint.ts
```diff
@@ -307,7 +312,7 @@ class NavigationEndpoint extends YTNode {
     if (parse && result)
       return Parser.parseResponse(result.data);
 
-    return this.#call(actions, client);
```
akkadaska, line 310: In some cases this would execute #call twice.

src/parser/classes/livechat/items/LiveChatPaidSticker.ts
```diff
@@ -42,7 +43,8 @@ class LiveChatPaidSticker extends YTNode {
     this.author_name_text_color = data.authorNameTextColor;
     this.sticker = Thumbnail.fromResponse(data.sticker);
     this.purchase_amount = new Text(data.purchaseAmountText).toString();
-    this.context_menu = new NavigationEndpoint(data.contextMenuEndpoint);
+    this.menu_endpoint = new NavigationEndpoint(data.contextMenuEndpoint);
+    this.context_menu = this.menu_endpoint;
```
akkadaska, line 47: Like others, this should be named menu_endpoint, not context_menu. To avoid breaking changes, I make both available.

LuanRT:
Awesome! I'll review it in a bit. Thank you for your work on this!

src/parser/youtube/LiveChat.ts
```diff
@@ -180,6 +190,31 @@ class LiveChat extends EventEmitter {
     return data.actions.array().as(AddChatItemAction);
   }
 
+  async getItemMenu(item: ChatItemHasMenuEndpoint): Promise<Menu> {
```
LuanRT, line 193: Anyway we can have this method return a new object? Here's a small gist to demonstrate what I mean:

> ./youtube/ItemMenu.ts
```ts
//...

class ItemMenu {
  constructor(data: ParsedResponse) {
    this.#page = data;
  }
  
  /**
   * This should find and select the correct menu item.
   * So "title" would be either the label, text or icon_type.
   * If an item is not matched then throw an error. If it is matched
   * then call its endpoint.
   */
  async selectItem(title: string): Promise<ActionsResponse> {
    // ...
  }
  
  /**
   * Returns original InnerTube response.
   */
  async page(): ParsedResponse {
    return this.#page;
  }
}
```

Then `getItemMenu` in `LiveChat` at line 193 would return an `ItemMenu` object, which should take care of selecting items and calling their endpoints. Ex:

> ./youtube/LiveChat.ts
```js
//...
const response = await item.menu_endpoint.call(this.#actions, undefined, true);
return new ItemMenu(response);
```

This would simplify usage quite a lot, let me know what you think. Thanks!

akkadaska: Thank you for the review. I have some worries.

1. When `ItemMenu.selectItem()` is called, to request to YouTube, we need to access LiveChat#actions. However, because `LiveChat#actions` is private member, we cannot acces it inside from `ItemMenu`. So some way to access is needed.
My idea is that LiveChat should have a method to call request to YouTube like call():
```typescript
class LiveChat {
  public async call(endpoint: NavigationEndpoint) {
    const response = await endpoint.call(this.#actions, undefined, true); 
    return response;
  }
}
```
and when `ItemMenu.selectItem()` is called, LiveChat.call() will be called inside from `ItemMenu.selectItem()`.

2. `LiveChatTextMessage.inline_action_buttons` and `LiveChatAutoModMessage.moderation_buttons` should be typed as ItemMenu, right?

3. Selecting item by its label, text or icon_type is simple especially deleting chat, but this may cause unstable works. `label` and `text` changes if another language is selected when `InnerTube.create()` is called. `icon_type` does not changes, but in some case, it is difficult to predict action from icon_type. For example, the icon_type for `hide this user` is `REMOVE_CIRCLE`.
So, I want additional way to call endpoint: receive endpoint instanses (currently typed `Button`) and select the endpoint by endpoint itself. I mean `LiveChat.selectButtonItem(button: Button)` is still needed.

LuanRT: > When ItemMenu.selectItem() is called, to request to YouTube, we need to access LiveChat#actions. However, because LiveChat#actions is private member, we cannot acces it inside from ItemMenu. So some way to access is needed.
My idea is that LiveChat should have a method to call request to YouTube like call():

You don't need to create a new method ~
I think simply sharing the current `Actions` instance from `LiveChat` with the `ItemMenu` class would do, ex:
```js
class ItemMenu {
  #page;
  #actions;

  constructor(data: ParsedResponse, actions: Actions) {
    this.#page = data;
    this.#actions = actions;
  }
  // ...
}
```

Then:
```js
//...
const response = await item.menu_endpoint.call(this.#actions, undefined, true);
return new ItemMenu(response, this.#actions);
```
----

> LiveChatTextMessage.inline_action_buttons and LiveChatAutoModMessage.moderation_buttons should be typed as ItemMenu, right?

Nope, since `ItemMenu` will be a wrapper around actual menus it shouldn't interfere with individual parser types.

----

>  So, I want additional way to call endpoint: receive endpoint instanses (currently typed `Button`) and select the endpoint by endpoint itself. I mean `LiveChat.selectButtonItem(button: Button)` is still needed.

Yes, I agree. For that, you can simply expose items from the menu as public properties in the `ItemMenu` class and add an additional method or make `ItemMenu#selectItem` accept either a string or an item of type `Button`.

akkadaska: OK! Thank you for your advice! Now I understand your idea.
I will work on to follow your design and request a review again.

akkadaska:
Now `LiveChat.getItemMenu(item)` returns a `Promise` of new class `ItemMenu`. 
We can select menu by `icon_type` of buttons like: 
```typescript
const itemMenu = await liveChat.getItemMenu(item);
itemMenu.selectItem('DELETE');
```
Also, we can get `Menu` instance by ItemMenu.menu(), which contains `Button` instances, and select item by `Button` instance like:
```typescript
const itemMenu = await liveChat.getItemMenu(item);
const targetButton = itemMenu.menu().items[0];
itemMenu.selectItem(targetButton);
```

I think we have one more task.
`LiveChatTextMessage.inline_action_buttons` or `LiveChatAutoModMessage.moderation_buttons` also contain `Button` instances to call. Currently, there are no appropriate way to call such `Button` items.
We can select item by `Button` using `ItemMenu.selectItem(button)`. However, to get `ItemMenu` instance, we need to call `liveChat.getItemMenu(item)`, which causes unnecessary request to YouTube.

My only idea is that `LiveChat` still should have a method to select item by `Button` instance. Please tell me your idea.

LuanRT:
> My only idea is that `LiveChat` still should have a method to select item by `Button` instance. Please tell me your idea.

I get what you're saying, the item menu is its own thing so that's why I suggested it to be a separate class. 

A generic button selector in the `LiveChat` class should be fine, unless we come up with a more elaborate way to deal with those additional buttons, but I can't really think of anything at the moment.

src/parser/youtube/ItemMenu.ts
```diff
@@ -0,0 +1,59 @@
...
+      });
+      if (!button) {
+        throw new InnertubeError('Button not found.');
+      }
+      endpoint = button.as(MenuServiceItem).endpoint;
+    }
+    if (!endpoint.metadata.api_url) {
+      throw new InnertubeError('Response did not have an "api_url" property. The call may have failed.');
+    }
+    const response = await this.#actions.execute(endpoint.metadata.api_url, {...endpoint.payload, parse: true});
```
LuanRT, line 45: One question, is it possible to use the button endpoint's own `callTest` method or does it error out? I feel like the use of execute here might be redundant.

Ex:
```js
//...
const response = await endpoint.callTest(this.#actions, { parse: true });

File: src/parser/youtube/ItemMenu.ts [hunk 6]
```diff
@@ -0,0 +1,59 @@
...
+  #actions: Actions;
+  #menu: Menu;
+  constructor(data: ParsedResponse, actions: Actions) {
+    this.#page = data;
+    this.#actions = actions;
+    const menu = data?.live_chat_item_context_menu_supported_renderers;
+    if (!menu || !menu.is(Menu)) {
+      throw new InnertubeError('Response did not have an "live_chat_item_context_menu_supported_renderers" property. The call may have failed.');
+    }
+    this.#menu = menu.as(Menu);
```
LuanRT, line 20: How about we expose only the `items` prop of the menu in this class? I think it would make more sense since it's already named `ItemMenu`. That way, instead of having to rely on the `getter` (e.g; `itemMenu.menu().items`), they could be accessed through `itemMenu.items` which is a cleaner in my opinion.

src/parser/youtube/ItemMenu.ts
akkadaska: I have no special reason to use `execute`, so I change it to use `callTest`. I confirmed it works well.

src/parser/youtube/ItemMenu.ts
akkadaska: Yours is better. I fixed it.

akkadaska:
Finally I added `LineChat.selectButton()` to select moderation buttons in `LiveChatTextMessage.inline_action_buttons` or `LiveChatAutoModMessage.moderation_buttons`. And I fixed to follow your review.

I think all tasks about #202 is completed and it's ready to merge. Please review again.

LuanRT (approved):
Great! Lgtm too!
I'll check the rest of the code to make sure we're not missing anything and then merge it ~ 
And again, thank you!

PR #251, opened 2022-12-01
Fix author and thumbnails for autogenerated playlists

absidue:
# Pull Request Template

## Description

While testing https://github.com/FreeTubeApp/FreeTube/pull/2903, we noticed that autogenerated playlists like https://m.youtube.com/playlist?list=RDCLAK5uy_nxNaxtS5HwGabWeGHsAADVA4KKl5l12eQ are partially supported by this library, with it spitting out this error:
casting_error

This pull request fixes that casting issue and also fixes the author field on the playlist being empty, the only missing bit of information is the thumbnails for the playlist author (empty array), although YouTube doesn't show it either, so I think that's fine.

The changes in this pull request work well with this FreeTube pull request https://github.com/FreeTubeApp/FreeTube/pull/2903 and were tested with it. Interestingly when I wanted to add a unit test to this library to test the autogenerated playlist, I ran into this error:
unit_test_error
Which is weird, because YouTube will happily return the playlist when the library is used inside of FreeTube. If I had to guess the differences are probably user agent related. Happy to add a unit test if there is a way to make it work in the unit test.

Here are the relevant lines of code in FreeTube, yes we don't have the most standard setup for YouTube.js 😬, see https://github.com/FreeTubeApp/FreeTube/pull/2855 for some of the reasoning behind our setup. We also disable CORS entirely in Electron, so that we don't have to deal with that.

[_scripts/webpack.renderer.config.js | lines 129-141](https://github.com/FreeTubeApp/FreeTube/blob/c98b6ab3f27895cd81dd302f17ef5473e9691c58/_scripts/webpack.renderer.config.js#L129-L141)

[src/renderer/helpers/api/local.js | lines 18-43](https://github.com/FreeTubeApp/FreeTube/blob/c98b6ab3f27895cd81dd302f17ef5473e9691c58/src/renderer/helpers/api/local.js#L18-L43)

[src/main/index.js | lines 175-183](https://github.com/FreeTubeApp/FreeTube/blob/c98b6ab3f27895cd81dd302f17ef5473e9691c58/src/main/index.js#L175-L183)

## Type of change

## Checklist:

- [ ] I have added tests that prove my fix is effective or that my feature works

### DIFF
--- a/src/parser/youtube/Playlist.ts
+++ b/src/parser/youtube/Playlist.ts
@@ -7,6 +7,7 @@ import VideoOwner from '../classes/VideoOwner';
 import PlaylistMetadata from '../classes/PlaylistMetadata';
 import PlaylistSidebarPrimaryInfo from '../classes/PlaylistSidebarPrimaryInfo';
 import PlaylistSidebarSecondaryInfo from '../classes/PlaylistSidebarSecondaryInfo';
+import PlaylistCustomThumbnail from '../classes/PlaylistCustomThumbnail';
 import PlaylistVideoThumbnail from '../classes/PlaylistVideoThumbnail';
 import PlaylistHeader from '../classes/PlaylistHeader';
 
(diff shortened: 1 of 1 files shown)


LuanRT:
Hey, thanks for taking the time to fix this!

>  Interestingly when I wanted to add a unit test to this library to test the autogenerated playlist, I ran into this error.

Just tried adding a unit test with the id `OLAK5uy_nGOuZnWRbeqZlch9DFLklW59Zrzyq2lx4` and it worked. So I am a bit confused, maybe there was something wrong with the id you tried?

absidue:
I used this ID `RDCLAK5uy_nxNaxtS5HwGabWeGHsAADVA4KKl5l12eQ` (see link in the description above), which works in the mobile YouTube app and FreeTube (with this YouTube.js PR but not with ytpl because that scrapes the html) but doesn't work on the desktop YouTube website or in the unit test.

LuanRT:
Alright, got it. Going to merge this now, we'll see about unit tests later.

PR #271, opened 2023-01-02
feat: Add is_live and is_upcoming to VideoDetails

absidue:
## Description

The current `is_live_content` property only tells you whether the video is a normal one or a live stream, it doesn't provide enough granularity on whether the live stream is live, upcoming or a live stream replay, it's true for all three of those. Extracting the `is_live` and `is_upcoming` properties as well, solves that.

Below is a table of the values of those 3 properties for the 4 types of videos mentioned above.

| Type | `is_live` | `is_live_content` | `is_upcoming` |
| --- | --- | --- | --- |
| currently live live stream | `true` | `true` | `false` |
| live stream replay | `false` | `true` | `false` |
| upcoming live stream | `false` | `true` | `true` |
| normal video | `false` | `false` | `false` |

I haven't added any tests as it'll be hard to find continously running live streams and upcoming live streams that never actually start.

## Type of change

## Checklist:

### DIFF
--- a/src/parser/classes/misc/VideoDetails.ts
+++ b/src/parser/classes/misc/VideoDetails.ts
@@ -13,7 +13,9 @@ class VideoDetails {
   view_count: number;
   author: string;
   is_private: boolean;
+  is_live: boolean;
   is_live_content: boolean;
+  is_upcoming: boolean;
   is_crawlable: boolean;
 
   constructor(data: any) {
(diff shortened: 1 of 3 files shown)


ChunkyProgrammer:
I believe this is a continuous livestream:
https://youtu.be/jfKfPfyJRdk

And we used this for a test in yt-channel-info but we might need a better example: 
https://youtu.be/O3cCYok_ukk

absidue:
Added test now.

LuanRT (approved):

LuanRT (approved):

PR #296, opened 2023-01-28
Allow checking whether a channel has optional tabs

absidue:
## Description

Certain YouTube channels don't have all tabs, depending on what type of content they post. When you are displaying the channel in a GUI, you may want to hide the missing tabs, like YouTube itself does. Currently the only to do that with YouTube.js is to try calling `await channel.getShorts()` and check the error message, unfortunately that also means that if the tab does exist, you will have fetched the content of that tab before the user has even switched to it.

This pull request adds getters to check if the Videos, Shorts, Live Streams and Community tabs exist, without making any requests. Those 4 tabs are the only ones I've notice missing in various combinations depending on the channel. I can add getters for the other tabs just in case, if you would like me to?

## Type of change

## Checklist:

### DIFF
--- a/src/core/TabbedFeed.ts
+++ b/src/core/TabbedFeed.ts
@@ -47,6 +47,10 @@ class TabbedFeed extends Feed {
     return new TabbedFeed(this.#actions, response.data, false);
   }
 
+  hasTabWithURL(url: string): boolean {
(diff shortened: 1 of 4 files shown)


LuanRT (changes requested):
Screenshot 2023-01-29 014203

Some channels from the explore page don't have a playlists tab. So there should be a getter for that too.

LuanRT (approved):

PR #342, opened 2023-03-08
feat(VideoInfo): support get by endpoint + more info

patrickkfkan:
Adds support for getting video info by endpoint in `getInfo()`.  In addition, add `playlist` and `autoplay` to `VideoInfo`. 

The result is that you can now obtain video info within the context of a list such as `Playlist` or `Mix`.  To illustrate, consider the following endpoint for a `Mix` item:

```
// mixEndpoint:
{
  "type": "NavigationEndpoint",
  "payload": {
    "videoId": "XHedu3KcPP0",
    "playlistId": "RDXHedu3KcPP0",
    "params": "OALAAQE%3D",
    "continuePlayback": true
  },
  ...
}

// Call `getInfo()` with mixEndpoint:

const info = await youtube.getInfo(mixEndpoint);

// Now you can get not only info about the video (as referenced by id 'XHedu3KcPP0') itself, but 
// also other items in the list.

// info.playlist:
{
  "id": "RDXHedu3KcPP0",
  "title": "Mix - ...",
  "author": ...,
  "contents": [
    {
      "type": "PlaylistPanelVideo",
      ...
    }
  ]
  ...
}

```

`info.playlist` mirrors what you would see on YouTube:

image

Now, say you want to play the last item in the list and get info on that:
```
const lastVideo = info.playlist.contents.pop();
const newInfo = await youtube.getInfo(lastVideo.endpoint);

// `newInfo.playlist` will now have updated contents based on the `selected` video.
// In the context of a Mix, there would be new items added with some older ones removed.
```

In addition, you can get the Autoplay video endpoint:
```
const endpoint = info.autoplay_video_endpoint;

// or obtain from `info.autoplay`
```
This points to the next item in the list or, if there are none (end of list reached), a video as determined by YouTube autoplay logic.

Of course, you can still call `getInfo()` with `video_id: string`. Apparently, there won't be `playlist` info in this case.

### DIFF
--- a/src/Innertube.ts
+++ b/src/Innertube.ts
@@ -72,7 +72,7 @@ class Innertube {
 
   /**
    * Retrieves video info.
-   * @param video_id - The video id.
+   * @param target - The video id or `NavigationEndpoint`.
    * @param client - The client to use.
    */
   async getInfo(target: string | NavigationEndpoint, client?: InnerTubeClient): Promise<VideoInfo> {
--- a/src/Innertube.ts
+++ b/src/Innertube.ts
@@ -31,7 +31,7 @@ import type Format from './parser/classes/misc/Format.js';
 import type { ApiResponse } from './core/Actions.js';
 import type { IBrowseResponse, IParsedResponse } from './parser/types/index.js';
 import type { DownloadOptions, FormatOptions } from './utils/FormatUtils.js';
-import { generateRandomString, throwIfMissing } from './utils/Utils.js';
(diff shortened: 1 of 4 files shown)


src/Innertube.ts
```diff
@@ -75,13 +75,45 @@ class Innertube {
    * @param video_id - The video id.
```
ChunkyProgrammer, line 75: Should probably update the comment here

patrickkfkan: Thanks for pointing this out. Updated.

LuanRT (approved):

PR #349, opened 2023-03-11
feat(yt): add support for movie items and trailers

patrickkfkan:
1. Add `GridMovie` and `HorizontalMovieList` parsers for movie-related elements (easy to find them in Explore -> Films)
2. Add `PlayerLegacyDesktopYpcTrailer` and `YpcTrailer` parsers for trailer info that appear in `errorScreen` of `/player` reponse, typically when viewing a non-purchased movie or film.
3. Add `getTrailerInfo()` and `has_trailer` getter to `VideoInfo`, which uses data from the parsers in (2) above. Usage:

```
const info = youtube.getInfo(video_id);

if (info.has_trailer) {
  const trailer_info = info.getTrailerInfo();  // Returns `VideoInfo` instance
  const format = trailer_info.chooseFormat(...);
  ...
}
```

### DIFF
--- a/src/parser/classes/GridMovie.ts
+++ b/src/parser/classes/GridMovie.ts
@@ -1,0 +1,34 @@
+import Parser from '../index.js';
+import Text from './misc/Text.js';
+import Thumbnail from './misc/Thumbnail.js';
+import NavigationEndpoint from './NavigationEndpoint.js';
+import { YTNode } from '../helpers.js';
+import MetadataBadge from './MetadataBadge.js';
+
+class GridMovie extends YTNode {
+  static type = 'GridMovie';
+
+  id: string;
+  title: Text;
+  thumbnails: Thumbnail[];
+  duration: Text | null;
+  endpoint: NavigationEndpoint;
+  badges: MetadataBadge[];
+  metadata: Text;
+  thumbnail_overlays;
+
+  constructor(data: any) {
+    super();
+    const length_alt = data.thumbnailOverlays.find((overlay: any) => overlay.hasOwnProperty('thumbnailOverlayTimeStatusRenderer'))?.thumbnailOverlayTimeStatusRenderer;
+    this.id = data.videoId;
+    this.title = new Text(data.title);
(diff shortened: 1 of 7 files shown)


src/parser/classes/GridMovie.ts
```diff
@@ -0,0 +1,34 @@
...
+  thumbnails: Thumbnail[];
+  duration: Text | null;
+  endpoint: NavigationEndpoint;
+  badges: MetadataBadge[];
+  metadata: Text;
+  thumbnail_overlays;
+
+  constructor(data: any) {
+    super();
+    const length_alt = data.thumbnailOverlays.find((overlay: any) => overlay.hasOwnProperty('thumbnailOverlayTimeStatusRenderer'))?.thumbnailOverlayTimeStatusRenderer;
```
absidue, line 22: Please use the parsed thumbnail_overlays array (see bottom of this constructor) instead of the raw data (ideally the only place in YouTube.js that should be handling raw data and have any mention of renderers, is the Parser itself).

Here is an example of how you can do it:

https://github.com/LuanRT/YouTube.js/blob/main/src/parser/classes/PlaylistVideo.ts#L56

absidue (changes requested):

src/parser/classes/GridMovie.ts
patrickkfkan: I actually mirrored this part from `GridVideo`. 

Type of `duration` is `Text` which I prefer for consistency with `GridVideo`. If I obtain `length_alt` through parsed `thumb_overlays`, then I would have to do something like this for setting value of `duration`:

```
const length_alt = this.thumbnail_overlays.firstOfType(ThumbnailOverlayTimeStatus);
this.duration = data.lengthText ? new Text(data.lengthText) : length_alt?.text ? new Text({ simpleText: length_alt.text }) : null;
```
I'm fine with either way. What do you think?

LuanRT: Yup, this is fine.

LuanRT (approved):

PR #351, opened 2023-03-12
feat!: Add support for OTF format streams

absidue:
## Description

Some videos have formats that are a bit different than other ones, YouTube seems to refer to these as OTF. According to NewPipeExtractor (https://github.com/TeamNewPipe/NewPipeExtractor/blob/dev/extractor/src/main/java/org/schabi/newpipe/extractor/services/youtube/DeliveryType.java#L23-L38) these can occur on videos with low view counts and live stream replays that have just been reencoded to normal videos. Instead of allowing you to request arbitrary chunks with the `range` query parameter, instead you use the `sq` query parameter to request sequences, the first sequence (`&sq=0`) contains the duration of the sequences, so we need to request it while building the DASH manifest.

This implementation was inspired by the NewPipeExtractor implementation, so big thanks to them :).

Here is a video with OTF streams: https://www.youtube.com/watch?v=DJ8GQUNUXGM
The 720p and 360p streams are normal, the other ones (144p, 240p and 480p) are OTF streams

## ⚠️ Breaking change ⚠️
The `toDash` functions are now asynchronous, they return a `Promise<string>` instead of a `string`, as we need to fetch the first sequence while building the manifest.

## Type of change

Please delete options that are not relevant.

## Checklist:

### DIFF
--- a/README.md
+++ b/README.md
@@ -166,7 +166,7 @@ const videoInfo = await youtube.getInfo('videoId');
 // now convert to a dash manifest
 // again - to be able to stream the video in the browser - we must proxy the requests through our own server
 // to do this, we provide a method to transform the URLs before writing them to the manifest
-const manifest = videoInfo.toDash(url => {
+const manifest = await videoInfo.toDash(url => {
   // modify the url
   // and return it
   return url;
--- a/src/parser/classes/misc/Format.ts
+++ b/src/parser/classes/misc/Format.ts
@@ -5,6 +5,7 @@ import type { RawNode } from '../../index.js';
 class Format {
   itag: number;
   mime_type: string;
+  is_type_otf: boolean;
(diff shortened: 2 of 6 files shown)


absidue:
Okay so apparently videos with OTF streams have the dash url field populated, so we don't need to generate these manifests in YouTube.js.

LuanRT:
I just tested this and also the dash manifest YouTube returns and both work great. This could be useful for those who need to filter out unwanted formats.

absidue:
Yep just remembered that FreeTube can't use YouTube's one because, videojs-http-streaming (the part of video.js that handles DASH) only supports MP4-DASH, it doesn't support WebM-DASH.

LuanRT (approved):

PR #366, opened 2023-03-22
fix(toDash): Generate unique Representation ids

absidue:
## Description

It looks like video.js changed something on their end since I added support for multiple audio tracks to the DASH manifest, now it seems to require unique ids for all representations, otherwise it just silently fails to switch audio tracks :(.

This was tested in FreeTube with video.js, hopefully this doesn't break shaka (I couldn't find anything in the spec that mandated that the ids consist of only digits, so I'm not expecting anything to break).

Here is a video that can be used as a test, it contains 3 English audio tracks, the main one, one with the interviewee dubbed and a third one with audio descriptions: https://youtu.be/Kn56bMZ9OE8

## Type of change

## Checklist:

### DIFF
--- a/src/utils/FormatUtils.ts
+++ b/src/utils/FormatUtils.ts
@@ -428,8 +428,15 @@ class FormatUtils {
     const url = new URL(format.decipher(player));
     url.searchParams.set('cpn', cpn || '');
 
+    let id;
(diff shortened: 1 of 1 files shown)


LuanRT (approved):
Tested with shaka and it worked correctly.

PR #388, opened 2023-04-24
refactor!: overhaul core classes and remove redundant code

LuanRT:
### What
To request video info, YouTube.js used to rely on a function from `Actions` called `getVideoInfo`. In fact, it was the only function that could make requests to `/player` in v1 - v2.

After a while, `Actions#execute` was implemented, thus rendering `getVideoInfo` redundant. In this PR, that function is removed and endpoints are organized into individual files so we don't have to declare a payload object every time we want to send a request.

### Additional Improvements
* `Player.ts` - Client version is now appended to deciphered URLs.
* `HTTPClient.ts` - Set `userAgent`, `osName`, `osVersion` and `platform` for Android clients.
* From now on, `userAgent` is only included in the payload of Android clients.
* Clients have been moved to a separate folder and to import them you must use the `Clients` namespace.
* Mixins are also in their own folder now and can only be imported through the `Mixins` namespace.


absidue:
As those classes contain purely static properties and have no state, they could be converted to const variables and functions instead. While they wouldn't get treeshaken, as they are referenced by other parts of YouTube.js that don't get treeshaken, it would still benefit projects that import YouTube.js and use a bundler, as they can inline functions and varialbes and they also tend not to mangle the names of class properties (javascript is very dynamic so renaming class properties could break stuff), but they'll happily mangle the names of functions and variables.

Just thought I would mention it, as this is a new feature, so making changes to this PR now wouldn't be breaking any existing functionality.

you could use a naming scheme like this (feel free to pick a completely different one of course):
variables: `<NAME>_ENDPOINT_PATH` e.g. `BROWSE_ENDPOINT_PATH`
functions: `build<Name>EndpointPayload` e.g. `buildBrowseEndpointPayload`

LuanRT:
@absidue 
Thank you for the suggestions.

I won't be going with pure functions such as `buildNotificationGetUnseenCountEndpoint` because it would be too verbose and difficult to maintain. I'm thinking of refactoring this to a single file called `Endpoints` and put everything there as objects.

Here's an example:
```ts
// ...

export const BrowseEndpoint = {
  PATH: "/browse" as const,
  /**
   * Builds a `/browse` request payload.
   * @param opts - The options to use.
   * @returns The payload.
   */
  build: (opts: BrowseEndpointOptions): IBrowseRequest => {
    return {
      ...{
        browseId: opts.browse_id,
        //....
      },
    };
  }
};
```

And since TypeScript 5.0 [supports the `export type * from 'mod'`](https://github.com/microsoft/TypeScript/pull/52217) syntax, we can just do something like this:
```ts
export * from './Endpoint.js'; // This only exports the objects.
export type * from './Endpoint.js'; // This exports the types.
```

Let me know if you have anything else in mind.

absidue:
Looks good to me 😄

LuanRT:
Should probably rename this to something else. I ended up refactoring way more than I first intended.

PR #393, opened 2023-04-30
fix(VideoInfo): Use microformat view_count when videoDetails view_count is NaN

absidue:
Looks like YouTube removed the view count from the video details, as I'm not sure if it's just on my machine or for everyone, it still tries to use the view count from the video details first before falling back to the microformat one. I wonder if this didn't show up earlier because we were using a hardcoded visitor ID?


LuanRT:
I think this happens because of the shorts params. If you manually send a request with the `Actions` class it still returns it in the video details object.

LuanRT (approved):

PR #397, opened 2023-05-06
fix(bundles): Use ESM tslib build for the browser bundles

absidue:
tslib is weird. The esbuild version that YouTube.js uses, isn't able to automatically pick the correct tslib build to bundle into the browser bundles, so it uses the commonjs one, which isn't treeshakable. This makes it pick the correct one, reducing the size of the bundles:

| Bundle | Before | After |
| --- | --- | --- |
| `browser.js` | `893.3kb` | `875.8kb` |
| `browser.min.js` | `479.6kb` | `471.7kb` |

More recent versions of esbuild added support for tslib's weirdness, however they now aggressively preserve comments while bundling (previously they only preserved webpack's magic comments), so that tools like vite can add magic comments without esbuild stripping them. Unfortunately that means that it blows up the sizes of the `node.cjs` and `browser.js` bundles, because newer esbuild versions keep **all** of the jsdoc comments 😔. The `browser.min.js` bundle isn't affected by that, as part of the mimification process is striping comments.
Doesn't look like there is a sensible way to disable that behaviour.


LuanRT:
Very strange behavior indeed. Thanks for the PR.

PR #404, opened 2023-05-16
Add support for YouTube Music mood filters

Buroni:
## Proposed changes
- Add `filters` and `applyFilter` methods to `ytmusic.HomeFeed`, which allows results to be filtered by mood.

## Testing

```js
import { Innertube } from "youtubei.js";

const yt = await Innertube.create();

const home = await yt.music.getHomeFeed();

console.log(home.filters); // [ 'Workout', 'Relax', 'Energize', 'Commute', 'Focus' ]

console.log(await home.applyFilter('Relax')) // instance of `HomeFeed`

```

## Notes

`HomeFeed.applyFilter` is almost exactly the same as `Search.applyFilter`. I'd like to avoid this code duplication by using an inherited base class in the `ytmusic` classes. I think this would also provide some structure to the classes in this directory, as they implement the same kinds of methods e.g. `has_continuation`, `getContinuation`, `applyFilter`, `filters`... However I didn't want to make an invasive change without discussion.

### DIFF
--- a/src/parser/ytmusic/HomeFeed.ts
+++ b/src/parser/ytmusic/HomeFeed.ts
@@ -39,8 +39,7 @@ class HomeFeed {
       return;
     }
 
-    const section_list = tab.content?.as(SectionList);
-    this.header = section_list?.header?.as(ChipCloud);
+    this.header = tab.content?.as(SectionList).header?.as(ChipCloud);
     this.#continuation = tab.content?.as(SectionList).continuation;
     this.sections = tab.content?.as(SectionList).contents.as(MusicCarouselShelf, MusicTastebuilderShelf);
   }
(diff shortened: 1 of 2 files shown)


LuanRT:
> `HomeFeed.applyFilter` is almost exactly the same as `Search.applyFilter`. I'd like to avoid this code duplication by using an inherited base class in the `ytmusic` classes. I think this would also provide some structure to the classes in this directory, as they implement the same kinds of methods e.g. `has_continuation`, `getContinuation`, `applyFilter`, `filters`... However I didn't want to make an invasive change without discussion.

This sounds great. But maybe it should be done in a separate PR? Also, there's already a class that does that for YouTube feeds (https://github.com/LuanRT/YouTube.js/blob/main/src/core/mixins/FilterableFeed.ts). But we don't use it for YouTube Music because there are a few differences that we'd have to account for.

Buroni:
@LuanRT Yep agreed it should be in a separate PR!

LuanRT (approved):

PR #421, opened 2023-06-28
fix(OAuth): client identity matching

patrickkfkan:
YouTube seems to have changed their Auth script which breaks  `OAuth#getClientIdentity()`. The Client Identity regex used to match against the body of the script is bound to fail. The result is blocking of the entire process while the regex tries futilely to find a match.

This PR fixes the Client Identity matching part. From what I see, the request payload for device code also seems to have changed. Instead of `model_name`, it's now `device_model`:

image

So I have included this change in the PR too.

### DIFF
--- a/src/core/OAuth.ts
+++ b/src/core/OAuth.ts
@@ -96,7 +96,7 @@ export default class OAuth {
       client_id: this.#identity.client_id,
       scope: Constants.OAUTH.SCOPE,
       device_id: Platform.shim.uuidv4(),
-      model_name: Constants.OAUTH.MODEL_NAME
+      device_model: Constants.OAUTH.MODEL_NAME
     };
 
     const response = await this.#session.http.fetch_function(new URL('/o/oauth2/device/code', Constants.URLS.YT_BASE), {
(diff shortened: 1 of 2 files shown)


absidue:
If it really is the regex itself that was causing the process to lock up, then moving it to worker threads is not solving the actual core problem, it's just working around the issue.
You need to rewrite the regex so that it doesn't suffer from catastrophic backtracking.
Moving it to worker threads is like building elaborate traps in your house to stop burglars, instead of just closing the front door that is unlocked and wide open.

patrickkfkan:
I was just thinking of guarding against unforeseeable cases where catastrophic backtracking may occur, but you are right that the regex itself should be worked on instead.

So would you mind reviewing the modified regex to see if it is foolproof enough to handle future YT changes, as in not locking up the process when no match is possible? My knowledge of this is rather limited so would appreciate some help here..

maciekr1234:
I can confirm indeed regex and or model_name was causing my app to stop responding.

LuanRT:
I'll merge this PR right now and publish a new version as this is quite urgent. I plan on doing a small refactor in this portion of the code soon, so I can look into fixing the unsafe regular exp, but anyone feel free to contribute if you can.

Thanks Patrick!

PR #431, opened 2023-07-06
feat(toDash): Hoist duplicates from Representation to AdaptationSet

absidue:
According to the DASH specification, `codecs`, `audioSampleRate`, `frameRate` and `AudioChannelConfiguration` are all allowed to appear at the `AdaptationSet` level not just the `Representation` one. We can use that to slightly reduce the size of the manifests by specifying it on the `AdaptationSet`, instead of each `Representation`, if the value would be identical for every `Representation` in that `AdaptationSet`.

Examples of situations that can be useful in:
1. The representations in WebM adaptation sets tend to all have the same codec e.g. `opus`, `vp9` and `vp9.2` (HDR), so it makes sense to hoist the `codecs` attribute
2. For 4k+ 60fps HDR videos the `vp9.2` streams all tend to keep the original `60fps` frame rates, so it makes sense to hoist the `frameRate` attribute

This doesn't massively reduce the size of the manifests, as the majority of the size is occupied by YouTube's very long streaming URLs. I do have an idea I want to try though, having a `BaseURL` element at the `AdaptationSet` level with a common prefix for the URLs in the `BaseURL`s in the `Represenation`, this will need lots of experimentation, as developer specified `URLTransformer` function could produce all sorts of wacky URLs. Another thing to watch out for is OTF formats, as they use SegmentTemplates instead of specifying a `BaseURL` but would still be affected by `BaseURL` elements on the AdaptationSet.
TL;DR: The streaming URLs are the biggest contributor to the manifest size, I have an idea to shrink them but will need a lot of experimentation so it doesn't break anything.


LuanRT:
Looks like this has some conflicts that need to be resolved. Anywho, I already reviewed it and it looks good!

LuanRT (approved):

absidue:
Fixed the conflicts, looks like I managed to create to create two conflicting pull requests, sorry about that.

LuanRT (approved):

PR #446, opened 2023-07-16
feat(toDash): Add option to include thumbnails in the manifest

absidue:
The DASH spec allows manifests to provide thumbnails, the most common use case for them is showing a preview when seeking with the seekbar. Players like shaka and DASH.js don't have a built-in UI for them, but they do parse them and provide APIs to get the thumbnail for a given timestamp, [`Player#getThumbnails`](https://shaka-player-demo.appspot.com/docs/api/shaka.Player.html#getThumbnails) and [`MediaPlayer#provideThumbnail`](https://cdn.dashjs.org/latest/jsdoc/module-MediaPlayer.html#provideThumbnail__anchor) respectively.

I decided to make this feature opt-in with a setting for the toDash function, as it requires making a few HEAD requests to get the content length, to approximate the bandwidth for the storyboards and it's not supported by all players/they don't provide a built-in UI for them. Sticking it behind an option allows people that can and want to use it, to use it without slowing the manifest generation down for everyone else.

I will create a separate pull request after this is merged (to avoid merge conflicts), to add captions/subtitles to the manifests, those require making requests too and it doesn't make sense to include both WebVTT and TTML captions, if you know your player only supports one of them.

### DIFF
--- a/src/core/mixins/MediaInfo.ts
+++ b/src/core/mixins/MediaInfo.ts
@@ -7,6 +7,8 @@ import { InnertubeError } from '../../utils/Utils.js';
 import type Format from '../../parser/classes/misc/Format.js';
 import type { INextResponse, IPlayerResponse } from '../../parser/index.js';
 import Parser from '../../parser/index.js';
+import type { DashOptions } from '../../types/DashOptions.js';
+import PlayerStoryboardSpec from '../../parser/classes/PlayerStoryboardSpec.js';
 
 export default class MediaInfo {
   #page: [IPlayerResponse, INextResponse?];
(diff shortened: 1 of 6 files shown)


LuanRT:
Haven't reviewed this completely yet but the `toDash` fn implementation should be go in `MediaInfo` so we can avoid all that duplicate code (plus the burden of having to modify all of them if say, we add another parameter).
https://github.com/LuanRT/YouTube.js/blob/7e5c3648c1f7bd0876ad38039d283c948f23d52c/src/core/mixins/MediaInfo.ts#L42-L44

absidue:
Unfortunately YouTube Kids doesn't provide any storyboads, so the storyboards properties are on the YouTube Music `TrackInfo` and YouTube `VideoInfo` classes, so we can't really move that up into MediaInfo as the properties it needs to read don't exist there. (well we could probably do it in pure javascript, as the properties would exist at runtime, but typescript doesn't like that kind of stuff).

absidue:
We'll have the same issue with captions, as both YouTube and YouTube Kids provide them but YouTube music doesn't, at least not on the `TrackInfo` object.

LuanRT:
Yes I am aware, but can't we just throw an error if someone tries using `MediaInfo#toDash` with that option set to `true` but the required data doesn't exist? At least we'd get to keep the function signature the same across all VideoInfo-like classes.

I really doubt anybody is going to use that option by accident, unless they're mixing youtube+youtube music+youtube kids.

LuanRT:
Also, you do have access to the `MediaInfo#page`, which contains the parsed response. Both `/player` and `/next` responses.

absidue:
Move the toDash implementation to just be in the MediaInfo class again, thanks for the suggestion!

LuanRT:
Perfect. Awesome work as always! Thank you.

LuanRT (approved):

PR #465, opened 2023-08-06
fix(Playlist): Only try extracting the subtitle for the first page

### LINKED ISSUE #464: 5.8.0 `header.subtitle` failed due to `header` being `undefined` in Playlist constructor
### Steps to reproduce

- Save following script into a file e.g. `script.js`
- Install `youtube.js` (I did this in FreeTube project folder)
- `node script.js`

```js
const { Innertube } = require('youtubei.js')

async function createInnertube(options = { withPlayer: false, location: undefined, safetyMode: false, clientType: undefined, generateSessionLocally: true }) {
  return await Innertube.create({
    enable_safety_mode: false,
    generate_session_locally: !!options.generateSessionLocally
  })
}

(async () => {
  const innertube = await createInnertube()
  let playlist = await innertube.getPlaylist('PLKkxnBwFOJGKYQo3EGYjE4FhJUbBq_2U5')

  const videos = playlist.items

  while (playlist.has_continuation) {
    playlist = await playlist.getContinuation()

    videos.push(...playlist.items)
  }

  console.info('Done!', videos.length)
})()

```

#### Original steps for FT dev
- Checkout latest development branch of https://github.com/FreeTubeApp/FreeTube (072066fbe1384ab09bdb1a3eca5f8e78d1cdb2ad at the time of writing)
- yarn && yarn dev
- Paste https://www.youtube.com/watch?v=Rm9PJNF3sE0&list=PLKkxnBwFOJGKYQo3EGYjE4FhJUbBq_2U5 to top, press enter or the arrow button

### Failure Logs

```shell
See Current behavior
```

### Expected behavior

No error thrown

### Current behavior

image

### Version

Default

### Anything else?

Using node 18.6.1 not 18.7.0 due to `https://github.com/nodejs/node/issues/48855` 

### Checklist

absidue:
Playlist continuations don't have the header so we should only try extracting the subtitle for the first playlist page.

closes #464
bug introduced in #458, which added support for the subtitle

### DIFF
--- a/src/parser/youtube/Playlist.ts
+++ b/src/parser/youtube/Playlist.ts
@@ -36,7 +36,7 @@ class Playlist extends Feed<IBrowseResponse> {
     this.info = {
       ...this.page.metadata?.item().as(PlaylistMetadata),
       ...{
-        subtitle: header.subtitle,
+        subtitle: header ? header.subtitle : null,
         author: secondary_info?.owner?.as(VideoOwner).author ?? header?.author,
         thumbnails: primary_info?.thumbnail_renderer?.as(PlaylistVideoThumbnail, PlaylistCustomThumbnail).thumbnail as Thumbnail[],
         total_items: this.#getStat(0, primary_info),


src/parser/youtube/Playlist.ts
```diff
@@ -36,7 +36,7 @@ class Playlist extends Feed<IBrowseResponse> {
     this.info = {
       ...this.page.metadata?.item().as(PlaylistMetadata),
       ...{
-        subtitle: header.subtitle,
+        subtitle: header ? header.subtitle : null,
```
ChunkyProgrammer, line 39: The rest of this file uses `header?.` instead
Ex:
can_share: header?.can_share,
can_delete: header?.can_delete,
is_editable: header?.is_editable,
privacy: header?.privacy

absidue: Using `header?.subtitle` would result in it returning `undefined`, which would change the type of the subtitle property from `Text | null` to `Text | null | undefined`. This could be mitigated by doing `header?.subtitle ?? null`, but due to YouTube.js targetting ES2016, it gets transpiled to the rather verbose: `(_b = header === null || header === void 0 ? void 0 : header.subtitle) !== null && _b !== void 0 ? _b : null,`.

I can change the type of `subtitle` to be `Text | undefined`, if you would prefer that over null, I definitely don't want both `null` and `undefined` being possible?

LuanRT (approved):

PR #494, opened 2023-08-31
feat(Session): Add on_behalf_of_user session option

jeremyBanks:
This adds support for manually specifying the `onBehalfOfUser` property in the session's `Context.user`, which specifies which channel to use if multiple users are associated with the logged-in account. This can be optionally specified when creating an `Innertube` instance in the same way as `account_id` is. I think this is only relevant if you're using a cookie session.

Profiles are identified with what might be called a "Page ID":  when I use the account switcher in the YouTube UI to switch profiles, it's the I see in a redirect URL like `https://www.youtube.com/signin?action_handle_signin=true&authuser=0&pageid=<THIS CHANNEL ID>`, where we can see it referred to as `pageid`.

### DIFF
--- a/src/core/Session.ts
+++ b/src/core/Session.ts
@@ -63,6 +63,7 @@ export interface Context {
   user: {
     enableSafetyMode: boolean;
     lockedSafetyMode: boolean;
+    onBehalfOfUser?: string;
   };
   thirdParty?: {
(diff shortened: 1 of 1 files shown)


LuanRT (approved):
Thank you!

Issue #422, opened 2023-06-28
Add /kids endpoints


ChipWolf:
### Describe your suggestion

`/youtubei/v1/kids/+` endpoints exist, some are quite useful, such as `update_blocklist` to add channels to child account blocklists from a parent account. I don't see the facility to call them from this library, this would be a good addition :)

I'm aware of the following endpoints from what I've found online:
```
- kids/clear_blacklist
- kids/get_available_kids_curators
- kids/get_kids_flow_data
- kids/get_parent_approved_content
- kids/get_selected_kids_curators
- kids/update_blacklist
- kids/update_parent_approved_content
- kids/update_selected_kids_curators
```

### Checklist

- [X] I am running the latest version.
- [X] I checked the documentation and found no answer.
- [X] I have searched the existing issues and made sure this is not a duplicate.
- [X] I have provided sufficient information.

PR #503, opened 2023-09-12
feat(Kids): Add `blockChannel` command to easily block channels

rsandbach:
Added new command and classes to support the Report / Block channel for kids

image

Small change to ToggleButton to allow parsing when there is now defaultIcon defined.


rsandbach:
This would be a start to #422.

absidue:
Probably better to put the kids only block channel in `src/core/clients/Kids.ts` instead of the interaction manager, as they presumably can't be used for YouTube or YouTube Music?

Please place your new kids only parser classes in the `src/parser/classes/ytkids` instead of `src/parser/classes/channels`.

rsandbach:
I went back and forth on this. My initial implementation was under Kids, but these particular actions/UI elements are only available within the YouTube web client when logged into an account that curates content for supervised accounts/kids profiles.

LuanRT:
Just skimming through the code and it looks good. Will do a proper review tomorrow or Tuesday.

LuanRT (approved):

ChipWolf:
> Probably better to put the kids only block channel in `src/core/clients/Kids.ts` instead of the interaction manager, as they presumably can't be used for YouTube or YouTube Music?
> 
> Please place your new kids only parser classes in the `src/parser/classes/ytkids` instead of `src/parser/classes/channels`.

This is nothing to do with ytkids. This is for supervised child accounts that are set to 'Explore', giving them access to yt proper.

PR #509, opened 2023-09-17
fix(build): Inline package.json import to avoid runtime erros

### LINKED ISSUE #508: Application fails when `package.json` is absent
### Steps to reproduce

My node application is bundled for deployment, therefore there’s no package.json file in the final build. https://github.com/LuanRT/YouTube.js/blob/68df3218580db10c9a0932c93ff2ce487526ff1e/src/platform/node.ts#L27
This assumes that the file is always present and crashes my app.

### Failure Logs

```shell
ENOENT: no such file or directory, open '/home/package.json'
    at Object.openSync (node:fs:592:3)
    at Object.readFileSync (node:fs:460:35)
    at Object.<anonymous> (/home/beta_amadeus/plugins/youtube.cjs:82:39138)
    at Module._compile (node:internal/modules/cjs/loader:1255:14)
    at Module._extensions..js (node:internal/modules/cjs/loader:1309:10)
    at Module.load (node:internal/modules/cjs/loader:1113:32)
    at Module._load (node:internal/modules/cjs/loader:960:12)
    at ModuleWrap.<anonymous> (node:internal/modules/esm/translators:165:29)
    at ModuleJob.run (node:internal/modules/esm/module_job:192:25)
    at async DefaultModuleLoader.import (node:internal/modules/esm/loader:
```

### Expected behavior

The library should continue working without package.json, as it doesn’t contain any essential information for it to function. In fact, putting an empty (with just `{}`) `package.json` in my home folder fixes the issue. But this is a bad solution. My app isn’t even in the home directory…

### Current behavior

The library crashes.

### Version

Default

### Anything else?

_No response_

### Checklist

Azarattum:
Fixes #508.

This PR inlines **only used** properties from `package.json` file on node platform during build time. This is achieve via [ts-transformer-inline-file](https://github.com/jirutka/ts-transformer-inline-file) typescript plugin and [ts-patch](https://github.com/nonara/ts-patch) to enable support for it. `build:esm` script now uses `tspc` instead of `tsc`.


absidue:
Definitely like the idea of injecting the values at build time, although I did notice one thing. Looks like `tspc` has an optional dependency on `typescript` v4, but this repo currently uses v5, does that mean `tspc` will use v5 or will it be a downgrade using the v4 version instead?

Azarattum:
`ts-patch` supports typescript 5.x, I believe it uses whatever version is installed in the project. `ts-transformer-inline-plugin` is just a transformer, so it shouldn't affect anything.

absidue:
Ah that sounds good.

LuanRT:
Looks like there's a peer dependency conflict going on.

PR #526, opened 2023-10-21
chore(MediaInfo): Throw helpful errors when calling toDash or download for live and Post-Live DVR videos

### LINKED ISSUE #518: 6.4.1 - Downloading a live stream always stops after a few chunks
### Steps to reproduce

```
const videoId = '...'; // Some livestream
const data = await youtube.getBasicInfo(videoId);
const stream = await data.download({
    type: 'audio', // same result for video or video+audio
    quality: 'best', 
    format: 'mp4',
});
const file = createWriteStream(`test.m4a`);
for await (const chunk of Utils.streamToIterable(stream)) {
    console.log('Write!');
    file.write(chunk);
}
```

Consistently after five 'Write!' logs, it stops and no more data is written to the file.

### Failure Logs

```shell
Write!
Write!
Write!
Write!
Write!
```

### Expected behavior

I expect the stream to keep up with the livestream and keep on yielding chunks whenever they become available.

### Current behavior

It consistently stops after five chunks. 

I'm totally unsure, but maybe the chunksize in the download function is too large (10Mb) for live streaming?

### Version

Default

### Anything else?

_No response_

### Checklist

absidue:
YouTube.js doesn't support generating DASH manifest for live and Post-Live-DVR videos, so now we'll throw an error that points people towards using the dash_manifest_url and hls_manifest_url that are already provided by YouTube.

YouTube.js also doesn't support downloading those videos, because they consist of 5 segment chunks that are all individual files, so they require stitching together with ffmpeg and can't just be written back to back into the same file.

closes #518

### DIFF
--- a/src/core/mixins/MediaInfo.ts
+++ b/src/core/mixins/MediaInfo.ts
@@ -46,10 +46,16 @@ export default class MediaInfo {
    * @returns DASH manifest
    */
   async toDash(url_transformer?: URLTransformer, format_filter?: FormatFilter, options: DashOptions = { include_thumbnails: false }): Promise<string> {
+    const playerResponse = this.#page[0];
+
+    if (playerResponse.video_details && (playerResponse.video_details.is_live || playerResponse.video_details.is_post_live_dvr)) {
+      throw new InnertubeError('Generating DASH manifests for live and Post-Live-DVR videos, is not supported. Please use the DASH and HLS manifests provided by YouTube in `streaming_data.dash_manifest_url` and `streaming_data.hls_manifest_url` instead.');
+    }
+
     let storyboards;
 
-    if (options.include_thumbnails && this.#page[0].storyboards?.is(PlayerStoryboardSpec)) {
-      storyboards = this.#page[0].storyboards;
+    if (options.include_thumbnails && playerResponse.storyboards?.is(PlayerStoryboardSpec)) {
+      storyboards = playerResponse.storyboards;
     }
 
     return FormatUtils.toDash(this.streaming_data, url_transformer, format_filter, this.#cpn, this.#actions.session.player, this.#actions, storyboards);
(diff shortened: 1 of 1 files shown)


src/core/mixins/MediaInfo.ts
```diff
@@ -46,10 +46,16 @@ export default class MediaInfo {
    * @returns DASH manifest
    */
   async toDash(url_transformer?: URLTransformer, format_filter?: FormatFilter, options: DashOptions = { include_thumbnails: false }): Promise<string> {
+    const playerResponse = this.#page[0];
+
+    if (playerResponse.video_details && (playerResponse.video_details.is_live || playerResponse.video_details.is_post_live_dvr)) {
+      throw new InnertubeError('Generating DASH manifests for live and Post-Live-DVR videos, is not supported. Please use the DASH and HLS manifests provided by YouTube in `streaming_data.dash_manifest_url` and `streaming_data.hls_manifest_url` instead.');
```
LuanRT, line 52:
```suggestion
      throw new InnertubeError('Generating DASH manifests for live and Post-Live-DVR videos is not supported. Please use the DASH and HLS manifests provided by YouTube in `streaming_data.dash_manifest_url` and `streaming_data.hls_manifest_url` instead.');
```

src/core/mixins/MediaInfo.ts
```diff
@@ -46,10 +46,16 @@ export default class MediaInfo {
    * @returns DASH manifest
    */
   async toDash(url_transformer?: URLTransformer, format_filter?: FormatFilter, options: DashOptions = { include_thumbnails: false }): Promise<string> {
+    const playerResponse = this.#page[0];
```
LuanRT, line 49: This should be in snake case.

src/core/mixins/MediaInfo.ts
```diff
@@ -83,6 +89,12 @@ export default class MediaInfo {
    * @param options - Download options.
    */
   async download(options: DownloadOptions = {}): Promise<ReadableStream<Uint8Array>> {
+    const playerResponse = this.#page[0];
+
+    if (playerResponse.video_details && (playerResponse.video_details.is_live || playerResponse.video_details.is_post_live_dvr)) {
+      throw new InnertubeError('Downloading is not supported for live and Post-Live-DVR videos, as they are split up into 5 second segments that are individual files, which require using a tool like ffmpeg to stitch them together, so they cannot be returned in a single stream.');
```
LuanRT, line 95:
```suggestion
      throw new InnertubeError('Downloading is not supported for live and Post-Live-DVR videos, as they are split up into 5 second segments that are individual files, which require using a tool such as ffmpeg to stitch them together, so they cannot be returned in a single stream.');
```

LuanRT (approved):

PR #529, opened 2023-10-28
feat: use `overrides` instead of `--legacy-peer-deps`

JellyBrick:
- Resolves the peer dependency conflict issue caused by #509.


LuanRT:
Thanks!

PR #537, opened 2023-11-09
feat(Channel): Support new about popup

absidue:
This pull request is a bit of a mess in some places (*cough* `Channel` and `ChannelTagline` classes) so any recommendations on how to clean it up are welcome.

YouTube is getting rid of the about tab on channels with a `C4TabbedHeader`, on the auto-generated game channels they don't seem to be making that change yet (they have an `InteractiveTabbedHeader`). On the auto-generated Gaming and Music channels the about tab seems to be gone completely with no replacement.

This pull request adds support for the new about popup and integrates it into the existing `has_about` getter and `getAbout()` method on the `Channel` class. As the nodes are different, it will require changes on the library users side to handle the new nodes, but at least the getter and the function are the same. Sometimes YouTube doesn't return the about tab, but returns a link to it in the tagline instead of the popup code, so this pull request supports both cases (plus the old one which is still needed on the auto-generated game channels and on the rare occasion that you still get the about tab).

Test channels:
`UCUQo7nzH1sXVpzL92VesANw`: DIY Perks (user channel)
`UCvWtix2TtWGe9kffqnwdaMw`: Rihanna - Topic (auto-generated artist channel, same as the user channel but with less information in the about popup)
`UCQvWX73GQygcwXOTSf_VDVg`: Minecraft - Topic (auto-generated game channel, still has an about tab)


LuanRT:
To fix the issue with conflicting node lists, you can update the `concatMemos` function to this:
```ts
export function concatMemos(...iterables: Array<Memo | undefined>): Memo {
  const memo = new Memo();

  for (const iterable of iterables) {
    if (!iterable) continue;
    for (const item of iterable) {
      // Update existing items.
      const memo_item = memo.get(item[0]);
      if (memo_item) {
        memo.set(item[0], [ ...memo_item, ...item[1] ]);
        continue;
      }

      memo.set(...item);
    }
  }

  return memo;
}
```

Not sure if doing this will cause problems elsewhere though, so you might want to test it by loading some different feeds that use the `Feed` class in FreeTube or another app that makes use of these features.

absidue:
Still needs updates to the check for a continuation in `Feed#has_continuation` and `Feed#getContinuationData`

absidue:
The failing test is for the channels tab that YouTube has removed, so it's unrelated to this pull request.

absidue:
This is ready for review now :)

LuanRT (approved):

PR #544, opened 2023-11-27
fear(parser): Add `DecoratedAvatarView`

iBicha:
Added `DecoratedAvatarView` as I was getting a warning.

Then I got a parsing error
```
ParsingError: Type mismatch, got DecoratedAvatarView expected ContentPreviewImageView.
```
Updating PageHeaderView seems to fix the error


absidue:
I'm guessing this was copied straight from the the generator?

The generator doesn't currently handle view models properly, which is why it didn't tell you that you should have a separate class for AvatarView. Also instead of manually parsing the array of images, please do it the same way that is done here: https://github.com/LuanRT/YouTube.js/blob/main/src/parser/classes/ContentPreviewImageView.ts#L13

iBicha:
> I'm guessing this was copied straight from the the generator?

Yes, as I'm fairly new to this codebase.
Thanks for the feedback

src/parser/classes/DecoratedAvatarView.ts
```diff
@@ -0,0 +1,86 @@
+import { YTNode } from '../helpers.js';
+import type { RawNode } from '../index.js';
+import { NavigationEndpoint } from '../nodes.js';
+
+export default class DecoratedAvatarView extends YTNode {
+  static type = 'DecoratedAvatarView';
+
+  avatar: {
+      avatar_view_model: {
```
iBicha, line 9: @absidue do you mean that avatar_view_model needs to be of type AvatarView?

LuanRT: Yes. Ideally, there should be an AvatarView node. Anything item that has "model" or "renderer" in its name should have its own individual module. The generator can only recognize renderers it seems.

src/parser/classes/DecoratedAvatarView.ts
```diff
@@ -0,0 +1,35 @@
...
+      avatar_view_model: AvatarView
+    };
+  a11y_label: string;
+  renderer_context: {
+      command_context: {
+        on_tap: {
+          innertube_command: NavigationEndpoint
+        }
+      }
+    };
```
absidue, line 19:
```suggestion
  on_tap_endpoint: NavigationEndpoint;
```

src/parser/classes/DecoratedAvatarView.ts
```diff
@@ -0,0 +1,35 @@
...
+      avatar_view_model: new AvatarView(data.avatar.avatarViewModel)
+    };
+    this.a11y_label = data.a11yLabel;
+    this.renderer_context = {
+      command_context: {
+        on_tap: {
+          innertube_command: new NavigationEndpoint(data.rendererContext.commandContext.onTap.innertubeCommand)
+        }
+      }
+    };
```
absidue, line 33:
```suggestion
    this.on_tap_endpoint = new NavigationEndpoint(data.rendererContext.commandContext.onTap.innertubeCommand);
```

src/parser/classes/DecoratedAvatarView.ts
```diff
@@ -0,0 +1,35 @@
+import { YTNode } from '../helpers.js';
+import type { RawNode } from '../index.js';
+import { NavigationEndpoint } from '../nodes.js';
```
absidue, line 3:
```suggestion
import NavigationEndpoint from './NavigationEndpoint.js';
```

absidue (commented):
Just a few small nitpicks so simplify the structure, other than that it looks good to me.

iBicha:
> Just a few small nitpicks so simplify the structure, other than that it looks good to me.

Thanks, I wanted to simplify but I didn't know how much the wrapper code needed to be faithful to the original model.
These changes definitely makes things better

LuanRT (approved):

PR #546, opened 2023-11-29
Fix array indexes in generated classes

dnicolson:
Before:
```
InnertubeError: DecoratedAvatarView not found!
This is a bug, want to help us fix it? Follow the instructions at https://github.com/LuanRT/YouTube.js/blob/main/docs/updating-the-parser.md or report it at https://github.com/LuanRT/YouTube.js/issues!
Introspected and JIT generated this class in the meantime:
class DecoratedAvatarView extends YTNode {

...
          sources: {
            0: {
              url: data.avatar.avatarViewModel.image.sources.0.url,
              width: data.avatar.avatarViewModel.image.sources.0.width,
              height: data.avatar.avatarViewModel.image.sources.0.height
            },
```

After:
```
          sources: {
            0: {
              url: data.avatar.avatarViewModel.image.sources[0].url,
              width: data.avatar.avatarViewModel.image.sources[0].width,
              height: data.avatar.avatarViewModel.image.sources[0].height
```

### DIFF
--- a/src/parser/generator.ts
+++ b/src/parser/generator.ts
@@ -435,6 +435,18 @@ export function toTypeDeclaration(inference_type: InferenceType, indentation = 0
   }
 }
 
+function arrayToKeyPathString(arr: string[]) {
+  let formattedString = '';
+  
+  for (let i = 0; i < arr.length; i++) {
+    const element = arr[i];
+    const isInteger = Number.isInteger(Number(element));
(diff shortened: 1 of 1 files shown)


absidue:
This doesn't fully fix the issue, as sources is created as an object instead of array. It's probably better if you add proper support for arrays, instead of abusing the fallback handlers.

PR #579, opened 2024-01-13
refactor(Playlist): Ignore `ContinuationItem` nodes from `SectionList#contents`

LuanRT:
This should address some issues where the library would fetch the wrong continuation or even empty continuations.

**NOTE**: The solution in https://github.com/LuanRT/YouTube.js/commit/987f50604a0163f9a07091ce787995c6f6fddb75 no longer applies after this PR is merged, as empty continuations were all in `SectionList#contents`.  If you had any code in place to handle that error, it is safe to remove it after this PR is merged.

## Other related changes
1. Added `style` to `PlaylistVideo`. This helps identifying if the video is just a recommendation (`PLAYLIST_VIDEO_RENDERER_STYLE_RECOMMENDED_VIDEO`) or part of the playlist. 

2. `Playlist#items` can only return `PlaylistVideo` nodes now. Its return type is`ObservedArray<PlaylistVideo>`.

### DIFF
--- a/src/parser/classes/PlaylistVideo.ts
+++ b/src/parser/classes/PlaylistVideo.ts
@@ -23,6 +23,7 @@ export default class PlaylistVideo extends YTNode {
   upcoming?: Date;
   video_info: Text;
   accessibility_label?: string;
+  style?: string;
 
   duration: {
     text: string;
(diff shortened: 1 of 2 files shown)


iBicha (approved):
LGTM, thanks for the quick fix!

absidue:
Looks like this broke support for shorts only playlists (they use `ReelItem`s instead of `PlaylistVideo`s), as this code assumes that playlists only contain `PlaylistVideo`s. It results in the following error being thrown in the `items` getter: `Error: Expected node of any type PlaylistVideo, got ReelItem`

As I don't know what the equivalent for the style attribute is for `ReelItem`s or if recommended ones even appear in playlists that only contain shorts, I won't be able to make a pull request to fix this myself (as discussed in the other thread video recommendations only show up with you are logged into YouTube).

Here is an example of a shorts only playlist (the auto-generated shorts playlist for the `@YouTubeMusic` YouTube channel): `UUSHStaiwu-FAgp_RC_tBiLh9A`

PR #593, opened 2024-02-04
Add support for `react-native` platform

nugmanoff:
Following PR adds support for `react-native` as a target platform. Thank you for making the library platform-agnostic and including the guide on how to add support for a new one – it helped a ton!

I've added some docs explaining what steps need to be taken on RN side in order for things to work.

There are some rough edges around typing that could easily be fixed. Decided to throw it up here first to get some feedback.

### DIFF
--- a/src/platform/react-native.ts
+++ b/src/platform/react-native.ts
@@ -1,0 +1,79 @@
+// React Native Platform Support
+import type { ICache } from '../types/Cache.js';
+import { Platform } from '../utils/Utils.js';
+import sha1Hash from './polyfills/web-crypto.js';
+import package_json from '../../package.json' assert { type: 'json' };
+import evaluate from './jsruntime/jinter.js';
+
+class Cache implements ICache {
+  #persistent_directory: string;
+  #persistent: boolean;
+
+  constructor(persistent = false, persistent_directory?: string) {
+    this.#persistent_directory = persistent_directory || '';
+    this.#persistent = persistent;
+  }
+
+  get cache_dir() {
+    return this.#persistent ? this.#persistent_directory : '';
+  }
+
+  #getStorage() {
+    const storage = new ((globalThis as any).mmkvStorage as any)({ id: 'InnertubeCache' });
+    return storage;
+  }
+
+  async get(key: string) {
+    const storage = this.#getStorage();
+    return storage.getBuffer(key)?.buffer;
+  }
+
+  async set(key: string, value: ArrayBuffer) {
+    const storage = this.#getStorage();
+    storage.set(key, new Uint8Array(value));
+  }
+
+  async remove(key: string) {
+    const storage = this.#getStorage();
+    storage.delete(key);
+  }
+}
+
+Platform.load({
+  runtime: 'react-native',
+  server: false,
+  info: {
+    version: package_json.version,
+    bugs_url: package_json.bugs.url,
+    repo_url: package_json.homepage.split('#')[0]
+  },
+  Cache: Cache,
+  sha1Hash,
+  uuidv4() {
+    if (globalThis.crypto?.randomUUID()) {
+      return globalThis.crypto.randomUUID();
+    }
+
+    // See https://stackoverflow.com/a/2117523
+    return '10000000-1000-4000-8000-100000000000'.replace(/[018]/g, (cc) => {
+      const c = parseInt(cc);
+      return (
+        c ^
+        (window.crypto.getRandomValues(new Uint8Array(1))[0] & (15 >> (c / 4)))
+      ).toString(16);
+    });
+  },
+  eval: evaluate,
+  fetch: globalThis.fetch,
+  Request: globalThis.Request,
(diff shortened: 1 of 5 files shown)


monokaijs:
Your work is awesome, however I think you should make mmkv optional. Forcing using mmkv make this module unable to be used along with react-native-debugger (since mmkv does not work without jsi). Request response from Innertube always complicate to debug, missing the help from network debugging and a good console viewer like react-native-debugger could be a big deal with us.

nugmanoff:
hey @monokaijs! thanks for the feedback – appreciate it! 
I am pretty newbie-ish when it comes to RN, so good to know the limitations of `mmkv` that you've mentioned; what alternatives would you suggest to use as a backing storage for cache to be used with this Youtube.js?

monokaijs:
How do you think about customizable cache adapter? Or allow caching customizations?
I used to try to use `youtubei.js` in my RN app. I think you should make Caching optional since we can implement our own cache mechanism, in this case, I see mmkv is just an option of developer, not the library itself.

absidue:
@monokaijs You can already pass your own implementation of the ICache class to YouTube.js' `InnerTube.create` method, this pull request just provides a default implementation, through the `UniversalCache` class.

I think this is fine as it is, providing a default implementation is good, if someone wants to use a different caching mechanism, they can always provide their own cache implementation.

src/platform/react-native.ts
```diff
@@ -0,0 +1,79 @@
...
+    this.#persistent_directory = persistent_directory || '';
+    this.#persistent = persistent;
+  }
+
+  get cache_dir() {
+    return this.#persistent ? this.#persistent_directory : '';
+  }
+
+  #getStorage() {
+    const storage = new ((globalThis as any).mmkvStorage as any)({ id: 'InnertubeCache' });
```
absidue, line 22: Might be worth throwing an error with an explanation when `mmkvStorage` is missing, that will prevent any cryptic errors further down the line.

monokaijs:
I see no need this implementation anymore. With React Native, after adding some polyfills and configure metro and builder to support YouTube.js, you can use this module without any hassle.
I've successfully implemented this module into my project with just some polyfills (which are also available in this PR). I think no big difference between ReactNative and browser platform except fetch function (which we do not have to proxy in ReactNative since security headers are modificable).

src/platform/react-native.ts
monokaijs: Agree

LuanRT:
Sorry for letting this get stale, I need some opinions on whether this is still needed or not. The last comment in this thread makes me think the browser implementation is enough, but I have no experience with react native to verify that.

I already reviewed this a while ago so I'll merge it once we come to a conclusion.

monokaijs:
I think Browser Implementation is enough.

nugmanoff:
Hey @LuanRT I *think* it might work with browser implementation directly by passing custom `Cache` implementation in `Innertube.create` call. But I think it more about the bigger picture of indicating that `YouTube.js` is supported and can be run on `react-native` and having an out-of-the box way to do that (i.e. providing default `Cache` implementation using react native specific libraries and necessary documentation).

I think the last documentation bit on glueing everything together on the React Native runtime side (using polyfills) is the most important thing to consider here. It took me quite some time to figure out how to properly pick and apply polyfill libraries on React Native side and it also includes some configs around bundler and babel. But now basically anyone who is looking for using `YouTube.js` can just install it as a dependency (and leverage the default `Cache` impl) and drop the polyfills setup file that is presented in the docs and have everything up and running in a few minutes. Without it one would need to figure out the Cache and polyfills for themselves (and might very well give up on the last one, thinking that it is not compatible with RN runtime at all).

tl;dr I think having this merged would clearly benefit `YouTube.js` by making the fact that it could be used and run on React Native more explicit and whole set up process more streamlined and drop-in, at the expense of having *almost* similar to web platform setup file (except for `Cache` implementation) – so almost no maintenance overhead. 

@monokaijs thank you for your thoughts on this one as well! while I do agree with you to some extent, I think the fact that you have discovered my fork because it was opened as PR in this repo and that helped you have `YouTube.js` running in your RN project only highlights the fact that why it is important to have this *explicit* support merged

src/platform/react-native.ts
nugmanoff: @absidue makes sense! I've just left a https://github.com/LuanRT/YouTube.js/pull/593#issuecomment-2150285749 and if this is the only issue that is blocking as per @LuanRT consideration – I will get it fixed asap

LuanRT:
Thanks! : )
I'll be merging this now. It should be available in v10 (which is going to be released later today, probably).

LuanRT (approved):

sgebr01:
@nugmanoff  I'm getting issues though with trying to sign in due to the event emitter (I'm guessing this has to do with the event-target polyfill). Did you ever come across this? "TypeError: Cannot read property 'prototype' of undefined"

PR #596, opened 2024-02-11
feat: add support of cloudflare workers runtime

ayoubqrt:
HI thank you for you work 😄 ,

I resolved two problems that are needed to make the lib work in cloudflare workers runtime:
- removed `credentials` http header which is not available
- changed the way that the global `fetch` function is passed to the `Platform` class

And I have implemented the cache for cloudflare workers


LuanRT (approved):

LuanRT:
Thanks!

PR #609, opened 2024-02-28
feat(parser): Support LockupView and it's child nodes

absidue:
Appears on https://www.youtube.com/@YouTube/podcasts and the podcast tabs on other channels too.

```js
import { Innertube } from './dist/src/platform/node.js';

const innertube = await Innertube.create();

const channel = await innertube.getChannel('UCBR8-60-B28hp2BmDPdntcQ'); // @YouTube

await channel.getPodcasts();
```


absidue:
The test failure is unrelated, it happens on with the latest commit on the main branch too.

LuanRT (approved):

Issue #623, opened 2024-03-30
9.1.0 Video "the following content is not available on this app" returned with Android


iBicha:
### Steps to reproduce

```
const yt = await Innertube.create();
const videoInfo = await yt.getInfo('v', 'ANDROID');
const dash = await videoInfo.toDash();
```
Then play the dash somewhere

### Failure Logs

```shell
N/A
```

### Expected behavior

Returns a dash to the video stream

### Current behavior

Returns a video that says "the following content is not available on this app"

### Version

Default

### Anything else?

Same as https://github.com/iv-org/invidious/issues/4498

### Checklist

- [X] I am running the latest version.
- [X] I checked the documentation and found no answer.
- [X] I have searched the existing issues and made sure this is not a duplicate.
- [X] I have provided sufficient information.

wukko:
all requests with `ANDROID` client get redirected to `aQvGIIdgFDM` video, hence the different "content not available" video:

<img width="432" alt="image" src="https://github.com/LuanRT/YouTube.js/assets/71202418/29cd50a2-37d3-4aa8-a089-b4c775067b4a">


can be worked around by using `YTMUSIC_ANDROID` client instead, but it's just the matter of time before youtube updates that one too.

absidue:
@iBicha The solution it to not use the Android client anymore (we should probably remove support for it from the library), switch to use one that actually works.

The android client now requires you to send along a valid DroidGuard result with the requests, which YouTube.js has no way to do, unless you want to run and Android device and extract that DroidGuard results from it for every single request you do.

absidue:
@iBicha For the code snippet you provided the WEB client, which is the default when you don't force anything else, works just fine. Just remove the `"ANDROID"` string from your code.

LuanRT:
> @iBicha For the code snippet you provided the WEB client, which is the default when you don't force anything else, works just fine, as long as you use a real video ID of course. Just remove the `"ANDROID"` string from your code.

After a few rounds of debugging it seems that if you send the player payload encoded as protobuf and without the DroidGuard bytes, you can still get video info and streaming data (but if you are not logged in, the streaming data formats will all start returning 403 after a while, same behavior we had before using those special parameters).

So I'm guessing we could keep the Android client for those that use the library with an account.

iBicha:
@LuanRT Amazing, that seems to work well!

LuanRT:
@iBicha Great, thanks for testing :). Closing this as completed for now.

[...]

PR #624, opened 2024-03-31
fix(PlayerEndpoint): Workaround for "The following content is not available on this app" (Android)

LuanRT:
To fix this, I modified the protobuf parameter we were using from:
```
{
    "1": {
        "2": 6
    },
    "59": 1
}
```
To:
```
{
    "1": {
        "1": 1
    },
    "59": 1
}
```

This is probably a temporary solution but it seems to work fine for now. 

Unless you can't run our JavaScript interpreter ([Jinter](https://github.com/LuanRT/Jinter)), I recommend using the WEB client (default)  for something more stable.

### Testing
Run the following code and try to play the streaming URL in a browser.
```ts
import { Innertube } from 'youtubei.js';

const yt = await Innertube.create();

const videoInfo = await yt.getBasicInfo('ksEYRaIpP7A', 'ANDROID');

console.info("Playback URL:", videoInfo.streaming_data.formats.find((stream) => stream.quality_label == '720p').url);
```

Related:
https://github.com/LuanRT/YouTube.js/issues/623, https://github.com/iv-org/invidious/issues/4498

### DIFF
--- a/src/core/endpoints/PlayerEndpoint.ts
+++ b/src/core/endpoints/PlayerEndpoint.ts
@@ -1,3 +1,4 @@
+import { encodeShortsParam } from '../../proto/index.js';
 import type { IPlayerRequest, PlayerEndpointOptions } from '../../types/index.js';
 
 export const PATH = '/player';
@@ -43,7 +44,7 @@ export function build(opts: PlayerEndpointOptions): IPlayerRequest {
       client: opts.client,
       playlistId: opts.playlist_id,
       // Workaround streaming URLs returning 403 or getting throttled when using Android based clients.
-      params: is_android ? '2AMBCgIQBg' : opts.params
+      params: is_android ? encodeShortsParam() : opts.params
     }
   };
 }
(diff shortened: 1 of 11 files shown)


src/core/endpoints/PlayerEndpoint.ts
```diff
@@ -43,7 +44,7 @@ export function build(opts: PlayerEndpointOptions): IPlayerRequest {
       client: opts.client,
       playlistId: opts.playlist_id,
       // Workaround streaming URLs returning 403 or getting throttled when using Android based clients.
-      params: is_android ? '2AMBCgIQBg' : opts.params
+      params: is_android ? encodeShortsParam() : opts.params
```
pukkandan, line 47: In yt-dlp, we were using using `{1: {2: 6}} -> CgIQBg==` without `59: 1`. As you discovered, changing it to `{1: {1: 1}} -> CgIQBg==` is working. Thanks for the finding!

But could you pls explain what the `59: 1` is for?

LuanRT: Hi! `59: 1` is so we can still get `storyboards` and a few other fields I can't remember right now.

pukkandan: hm.. we are getting storyboards without it

PR #638, opened 2024-04-14
[Attempt] Fix for shelf type in Library

iBicha:
Attempt to fix #637 
From what I can see, only the history and playlists have ids (`library-recent` and `library-playlists-shelf`)
I don't know how to identify `WATCH_LATER`,  `LIKE`, or `CONTENT_CUT`. Those sections will still return undefined when accessed from their respective properties

PR #655, opened 2024-04-29
Item section target_id fix

iBicha:
A fix that was made in #638 but PR is closed


LuanRT (approved):

absidue:
@iBicha Please use the correct format for your pull request titles and commit messages so that they actually appear in the changelog (notice how your feature request in the previous release is missing from the changelog). YouTube.js uses the conventional commits standard.

As this pull request has already been merged I'll add an override, but please use the correct format when opening pull requests in the future.

PR #682, opened 2024-07-09
fix(Player): Fix extracting the n-token decipher algorithm

absidue:
This pull request adds support for the player `b22ef6e7`, which uses `String.prototype.split` and `Array.prototype.join`.

A Jinter update will be required to add support for `Array`.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -12,7 +12,7 @@
       ],
       "license": "MIT",
       "dependencies": {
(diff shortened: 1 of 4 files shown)


dylnuge:
Opened a PR to add the needed support to Jinter: https://github.com/LuanRT/Jinter/pull/13

LuanRT (approved):

Issue #711, opened 2024-07-30
<current version as of July 30th> <Songs stop>


wolfles:
### Steps to reproduce

play a song
wait

### Failure Logs

```shell
Unhandled Rejection DiscordRESTError: Invalid Form Body on POST /api/v10/applications/1265898592439107597/commands name: String value did not match validation regex. at RequestHandler.request (D:\SteamLibrary\steamapps\common\Bot Maker For Discord\node_modules\oceanic.js\dist\lib\rest\RequestHandler.js:96:15) at RequestHandler.authRequest (D:\SteamLibrary\steamapps\common\Bot Maker For Discord\node_modules\oceanic.js\dist\lib\rest\RequestHandler.js:81:21) at RESTManager.authRequest (D:\SteamLibrary\steamapps\common\Bot Maker For Discord\node_modules\oceanic.js\dist\lib\rest\RESTManager.js:47:29) at Applications.createGlobalCommand (D:\SteamLibrary\steamapps\common\Bot Maker For Discord\node_modules\oceanic.js\dist\lib\routes\Applications.js:78:30) at ClientApplication.createGlobalCommand (D:\SteamLibrary\steamapps\common\Bot Maker For Discord\node_modules\oceanic.js\dist\lib\structures\ClientApplication.js:54:46) at D:\SteamLibrary\steamapps\common\Bot Maker For Discord\bot.js:1537:30 at Array.forEach () at registerCommands (D:\SteamLibrary\steamapps\common\Bot Maker For Discord\bot.js:1535:19) at process.processTicksAndRejections (node:internal/process/task_queues:95:5) { code: 50035 }
play (@#0) >>> Play YouTube Song (@#2) >>> Error: Error: Failed to extract n-token decipher algorithm
PlayerError: Failed to extract n-token decipher algorithm at Player.extractNSigSourceCode (D:\SteamLibrary\steamapps\common\Bot Maker For Discord\node_modules\youtubei.js\bundle\node.cjs:17843:13) at Player. (D:\SteamLibrary\steamapps\common\Bot Maker For Discord\node_modules\youtubei.js\bundle\node.cjs:17720:28) at Generator.next () at fulfilled (D:\SteamLibrary\steamapps\common\Bot Maker For Discord\node_modules\tslib\tslib.js:166:62) at process.processTicksAndRejections (node:internal/process/task_queues:95:5)
```

### Expected behavior

play the whole song

### Current behavior

[...]

PR #713, opened 2024-07-31
refactor(Player): Generate and parse player script's `AST`

LuanRT:
Notes:
- The Syntax Tree is generated by [Jinter](https://github.com/LuanRT/Jinter) (which is built on top of [Acorn](https://github.com/acornjs/acorn)).
- While doing this may be slightly slower than using a regular exp, it is much more reliable (plus we already cache the player functions anyway).
- `findFunction` expects the functions to be inside a self-invoking function.

#### `Utils#findFunction` Options & Usage Example:
```ts
export type FindFunctionArgs = {
  /**
   * The name of the function.
   */
  name?: string;

  /**
   * A string that must be included in the function's code for it to be considered.
   */
  includes?: string;

  /**
   * A regular expression that the function's code must match.
   */
  regexp?: RegExp;
};
````

```ts
const source = '(function() {var foo, bar; foo = function() { console.log("foo"); }; bar = function() { console.log("bar"); }; })();';
const result = findFunction(source, { name: 'bar' });
// Or: const result = findFunction(source, { includes: 'console.log("bar")' });
console.log(result);
```
Output:
```
{
  start: 69,
  end: 110,
  name: 'bar',
  node: { /**/ },
  result: 'bar = function() { console.log("bar"); };'
}
```

---

Closes https://github.com/LuanRT/YouTube.js/issues/711

### DIFF
--- a/src/core/Player.ts
+++ b/src/core/Player.ts
@@ -1,5 +1,5 @@
 import { Log, LZW, Constants } from '../utils/index.js';
-import { Platform, getRandomUserAgent, getStringBetweenStrings, PlayerError } from '../utils/Utils.js';
+import { Platform, getRandomUserAgent, getStringBetweenStrings, findFunction, PlayerError } from '../utils/Utils.js';
 import type { ICache, FetchFunction } from '../types/index.js';
 
 const TAG = 'Player';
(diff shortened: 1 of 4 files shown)


src/core/Player.ts
```diff
@@ -223,13 +223,19 @@ export default class Player {
   static extractNSigSourceCode(data: string): string | undefined {
     const match = data.match(/b=(?:a\.split\(|String\.prototype\.split\.call\(a,)\n?(?:""|\("",""\))\).*?\}return (?:b\.join\(|Array\.prototype\.join\.call\(b,)\n?(?:""|\("",""\))\)\}/s);
```
LuanRT, line 224: Do we keep this here or should we always use the ast walker?

absidue: I think we should always use the AST walker.

PR #729, opened 2024-08-11
fix(Session): PoToken not being set correctly

cjh980402:
This is fir for missing potoken in player request's payload.
After that fix, then potoken is works so I don't see 'sign in~bot' error.


LuanRT:
Thanks :)!

PR #736, opened 2024-08-18
Fix account_index

### LINKED ISSUE #735: The `account_index` parameter does not work
### Steps to reproduce

1. Navigate to https://www.youtube.com/playlist?list=WL while logged into two accounts
2. Attempt to fetch the Watch Later playlist from the secondary account

```javascript
const youtube = await Innertube.create({
  cookie: document.cookie,
  account_index: 1,
  fetch: (...args) => fetch(...args),
})

const playlist = await youtube.getPlaylist('WL')
const videoIds = playlist.videos.map((video) => video.id)
console.log('Video IDs', videoIds)
```

In most cases, it was required to remove these lines:
https://github.com/LuanRT/YouTube.js/blob/4b60b97132b0ee42b41838f3336c582a7f7f7aec/src/utils/HTTPClient.ts#L75-L76

To prevent this response from `/playlist` requests:
```
)]}'
{"reload":"now"}
```

### Failure Logs

```shell
The playlist does not exist.
```

### Expected behavior

The playlist contents will be retrieved from the secondary account.

### Current behavior

The `X-Goog-AuthUser` header doesn't appear to indicate the active account as described in #229 and https://github.com/yt-dlp/yt-dlp/commit/34917076ad9844eddfa4ea97656d81a7fefe5d59.

Looking at the source of [desktop_polymer.js](https://www.youtube.com/s/desktop/f722e545/jsbin/desktop_polymer.vflset/desktop_polymer.js), the `X-Goog-AuthUser` header comes from the `SESSION_INDEX` value:
```
{
    Authorization: "AUTHORIZATION",
    "X-Goog-EOM-Visitor-Id": "EOM_VISITOR_DATA",
    "X-Goog-Visitor-Id": "SANDBOXED_VISITOR_ID",
    "X-Youtube-Domain-Admin-State": "DOMAIN_ADMIN_STATE",
    "X-Youtube-Chrome-Connected": "CHROME_CONNECTED_HEADER",
    "X-YouTube-Client-Name": "INNERTUBE_CONTEXT_CLIENT_NAME",
    "X-YouTube-Client-Version": "INNERTUBE_CONTEXT_CLIENT_VERSION",
    "X-YouTube-Delegation-Context": "INNERTUBE_CONTEXT_SERIALIZED_DELEGATION_CONTEXT",
    "X-YouTube-Device": "DEVICE",
    "X-Youtube-Identity-Token": "ID_TOKEN",
    "X-YouTube-Page-CL": "PAGE_CL",
    "X-YouTube-Page-Label": "PAGE_BUILD_LABEL",
    "X-YouTube-Variants-Checksum": "VARIANTS_CHECKSUM",
    "X-Goog-AuthUser": "SESSION_INDEX",
    "X-Goog-PageId": "DELEGATED_SESSION_ID"
}
```

Switching between accounts does not change the `ytcfg.data_.SESSION_INDEX` value, it is always `0`. A search on GitHub only found code using the `SESSION_INDEX` value or hard coding `0` or `1`.

### Version

Default

### Anything else?

This code successfully gets the contents of the Watch Later playlist of the active account:

```javascript
const youtube = await Innertube.create({
  cookie: document.cookie,
  fetch: (input, init = {}) => {
    const headers = new Headers()
    headers.append('X-YouTube-Client-Name', window.ytcfg.data_.INNERTUBE_CONTEXT_CLIENT_NAME)
    headers.append('X-YouTube-Client-Version', window.ytcfg.data_.INNERTUBE_CONTEXT_CLIENT_VERSION)
    headers.append('X-YouTube-Device', window.ytcfg.data_.DEVICE)
    headers.append('X-YouTube-Identity-Token', window.ytcfg.data_.ID_TOKEN)
    const options = { ...init, headers }
    return fetch(input, options)
  },
})

const response = await youtube.session.http.fetch(`/playlist?list=WL&pbj=1`, {
  baseURL: 'https://www.youtube.com',
})
const json = await response.json()

const playlist =
  json.response.contents.twoColumnBrowseResultsRenderer.tabs[0].tabRenderer.content.sectionListRenderer.contents[0]
    .itemSectionRenderer.contents[0].playlistVideoListRenderer

console.log('Video IDs', playlist.contents.map((video) => video.playlistVideoRenderer.videoId))
```

### Checklist

dnicolson:
This fixes #735.

The `X-Goog-PageId` header determines the active account, not the `X-Goog-Authuser` header. Data sync IDs come from the `/getDatasyncIdsEndpoint` endpoint and are saved in local storage.

The `prettyPrint` and `alt` parameters need to be removed to prevent this response to `/playlist` requests:
```
)]}'
{"reload":"now"}
```

Here are examples of data sync IDs for the number of accounts signed in.

Three accounts:
id_1||id_2::yt-player::yt-player-lv
id_0||::yt-player::yt-player-lv

Two accounts:
id_1||id_0::yt-player::yt-player-lv
id_0||::yt-player::yt-player-lv

One account:
id_0||::yt-player::yt-player-lv

### DIFF
--- a/src/utils/HTTPClient.ts
+++ b/src/utils/HTTPClient.ts
@@ -1,5 +1,13 @@
 import * as Constants from './Constants.js';
 
+declare global {
+  interface Window {
+    ytcfg: {
+      get(key: string): string;
+    };
+  }
+}
+
 import {
   Platform,
   generateSidAuth,
@@ -72,8 +80,10 @@ export default class HTTPClient {
       request_headers.set('Origin', request_url.origin);
     }
 
-    request_url.searchParams.set('prettyPrint', 'false');
-    request_url.searchParams.set('alt', 'json');
+    if (typeof window === 'undefined') {
+      request_url.searchParams.set('prettyPrint', 'false');
+      request_url.searchParams.set('alt', 'json');
+    }
 
     const content_type = request_headers.get('Content-Type');
 
@@ -142,7 +152,23 @@ export default class HTTPClient {
 
         if (sapisid) {
           request_headers.set('Authorization', await generateSidAuth(sapisid));
-          request_headers.set('X-Goog-Authuser', this.#session.account_index.toString());
+
+          if (typeof window !== 'undefined' && window.localStorage && this.#session.client_name === 'WEB') {
+            const dataSyncIds = [ window.ytcfg.get('DELEGATED_SESSION_ID') ];
+            const dataSyncIdsKeyExpression = /^(\d+)(?:\|\|(\d+))*(?:::yt-player::yt-player-lv)$/;
+            for (let i = 0; i < localStorage.length; i++) {
+              const key = localStorage.key(i);
+              const matches = key?.match(dataSyncIdsKeyExpression);
+              if (matches) {
+                dataSyncIds.push(...matches.slice(1).filter((id) => id !== dataSyncIds[0]));
+              }
+            }
+
+            request_headers.set('X-Goog-AuthUser', window.ytcfg.get('SESSION_INDEX'));
+            if (dataSyncIds[this.#session.account_index]) {
+              request_headers.set('X-Goog-PageId', dataSyncIds[this.#session.account_index]);
+            }
+          }
         }
 
         request_headers.set('Cookie', this.#cookie);


absidue:
This pull request seems to assume that every user of the library is only using it directly on YouTube's own website and this will blow up for everyone that doesn't (which is probably most users).

A better approach would be to add an extra parameter so the session creation so you can pass the relevant parameters into the library, the same way that visitor data and poToken's are handled.

dnicolson:
The changes only apply if the cookie is set, and this only works in the browser:
https://github.com/LuanRT/YouTube.js/blob/4942992/src/utils/HTTPClient.ts#L140

Are you suggesting limiting the changes to accepting a data sync ID and setting the `X-Goog-PageId` header? In that case, the `account_index` option would need to be removed entirely.

dnicolson:
I added a `client_name` check to make the intent clearer.

absidue:
> Are you suggesting limiting the changes to accepting a data sync ID and setting the X-Goog-PageId header?

Better that it breaking for people passing cookies to YouTube.js on Node, Deno or ReactNative or people using it in web browsers but not on the YouTube website.

dnicolson:
It should be more robust now, the only risks I see are:
- Only the web client will get the `X-Goog-AuthUser` header, which didn't appear to work previously.
- Other websites will not have additional URL search parameters.

Works with the web client and in Node.js.

PR #737, opened 2024-08-19
fix(ThumbnailOverlayResumePlayback): Update `percent_duration_watched` type

dnicolson:
PixelSnap 2024-08-19 at 22 17 08@2x


LuanRT:
Thanks!

Issue #748, opened 2024-09-11
What could it be 'InnertubeError: Streaming data not available' only in remote server but working locally?


rogerpadilla:
### Question

Hello, what could it be that my exact code works fine locally but when deploy it in remote servers I got this weird error: 

```
InnertubeError: Streaming data not available
    at Module.chooseFormat (/opt/src/node_modules/youtubei.js/src/utils/FormatUtils.ts:135:11)
    at VideoInfo.chooseFormat (/opt/src/node_modules/youtubei.js/src/core/mixins/MediaInfo.ts:150:24)
    at chooseFormat (/opt/src/server/helpers/yti.helper.ts:59:15)
```

My current related code is as following (compacted):
```ts
  /* I change this every half an hour automatically */
  let _yt: Innertube & { variabilityCreatedAt?: Date };

  async function getYt() {
    const randomMilliseconds = getRandomMilliseconds();
    const thresholdDate = new Date(Date.now() - randomMilliseconds);

    if (thresholdDate <= _yt?.variabilityCreatedAt) {
      return _yt;
    }

    const { poToken, visitorData } = await scrapeYtSession();

    _yt = await Innertube.create({
      po_token: poToken,
      visitor_data: visitorData,
      generate_session_locally: true,
      cache: new UniversalCache(false),
    });

    _yt.variabilityCreatedAt = new Date();

    return _yt;
  }

  async function getDownloadUrl(url: string) {
    const id = getVideoId(url);
    const yt = await getYt();
    const info = await yt.getBasicInfo(id);
    const format = chooseFormat(info);
    const decipheredUrl = format.decipher(yt.session.player);
    // logger.debug('** decipheredUrl', decipheredUrl);
    return decipheredUrl;
  }

  function chooseFormat(info: VideoInfo) {
    return info.chooseFormat?.({ type: 'video+audio', quality: 'bestefficiency' }) ?? info.streaming_data?.formats[0];
  }
```

thanks in advance.

### Checklist

- [X] I am running the latest version.
- [X] I checked the documentation and found no answer.
- [X] I have searched the existing issues and made sure this is not a duplicate.
- [X] I have provided sufficient information.

whiskeycola:
check `info.playability_status` maybe you detected a bot

rogerpadilla:
hey @whiskeycola, thanks for your help.

Yes, it seems you are right, below is what I get local vs remote server.

[...]

PR #756, opened 2024-09-18
feat: Add `WEB_EMBEDDED` client

unixfox:
This PR adds the `WEB_EMBEDDED_PLAYER` client.

This client is now essential for using the po_token technique since WEB client does not work anymore. See here: https://github.com/iv-org/invidious/pull/4923#issue-2529459521

This client is configured with the embedded URL configured to `"embedUrl":"https://www.google.com/"` in order to allow even some video that disallowed embedding on other websites. Not all videos, see comment: https://github.com/iv-org/invidious/pull/4923#issuecomment-2354848656


unixfox:
I'm also going to add WEB_CREATOR client since @rustynail1984 told me that it worked for him with po_token too: https://github.com/LuanRT/YouTube.js/issues/748#issuecomment-2348418093

PR #812, opened 2024-11-20
refactor(parser)!: Implement endpoint/command parsers

### LINKED ISSUE #810: Implement a proper endpoint/command mapper
### Describe your suggestion

Currently, the `Actions` class is doing all the heavy lifting when an endpoint's payload needs to be modified, and it is not very efficient.

### Other details

YouTube's web app uses an object that maps all endpoints/commands to their respective classes, which are responsible for providing API paths, building payloads, and more. We could do the same.

### Checklist

LuanRT:
Closes #810

The old request builders (under the `Endpoints` namespace) have been removed. Those who relied on them should now use the `NavigationEndpoint` node to build requests.

Example:
```ts
const continuationCommand = new YTNodes.NavigationEndpoint({
    continuationCommand: {
        request: 'CONTINUATION_REQUEST_TYPE_WATCH_NEXT',
        token: continuation
    }
});

const response = await continuationCommand.call(innertube.session.actions, /** additional payloads, etc */);
console.log(response);
```

### Why?
The old request builders were messy and incompatible with the parser. With this PR, the parser can correctly identify endpoints, commands, and build request payloads.


absidue:
Before this gets merged, I would like to try adding support for this gigantic refractoring in FreeTube, we definitely need to make changes on our side. I want to check that FreeTube can still do everything that it currently does, so that I can request changes if necessary before this is merged.

src/types/StreamingInfoOptions.ts
```diff
@@ -26,5 +26,5 @@ export interface StreamingInfoOptions {
    *
    * Defaults to `(audio_track_display_name) => audio_track_display_name + " (Stable Volume)"`
    */
-  label_drc_mutiple?: (audio_track_display_name: string) => string;
+  label_drc_multiple?: (audio_track_display_name: string) => string;
```
LuanRT, line 29: ^^ Technically, this is also a breaking change. Since this was going to be a big refactor anyway, I fixed some typos and removed deprecated code (like the YouTube Studio stuff).

absidue:
As this will trigger a major version bump because of the breaking changes, should we also use it as an opportunity to get rid of some of the dead nodes?

In a separate pull request of course (wouldn't want to make this one even bigger) but in the same release because it will also be a breaking change (TypeScript users will probably get type errors after the dead nodes are removed).

bcomnes:
Any pointers on how to refactor the onsie request to use this new class?

https://github.com/LuanRT/googlevideo/blob/96004ce991ac7846ab6a35b29203ae208a80413b/examples/onesie-request/main.ts#L73-L77

```js
...Endpoints.PlayerEndpoint.build({ ... })
```

LuanRT:
@bcomnes 
Something like this: https://github.com/LuanRT/YouTube.js/blob/main/src/Innertube.ts#L108-L118

bcomnes:
Hrmmm. afraid I'm still not quite getting it. 

Any plans to update the onesie example to use v12 of youtubei.js? I can reference that if you do.

LuanRT:
@bcomnes https://github.com/LuanRT/googlevideo/commit/7a106dc5f47a773ec989fbe49e84b0b7075dea69

PR #832, opened 2024-12-09
fix(Player): Fix signature algorithm extraction

absidue:
There is probably a cleaner way to handle this but this gets it working.

### DIFF
--- a/src/core/Player.ts
+++ b/src/core/Player.ts
@@ -224,14 +224,21 @@ export default class Player {
   }
 
   static extractSigSourceCode(data: string): string {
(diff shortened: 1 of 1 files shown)


bcomnes:
This fixes the `YOUTUBEJS][Player]: Failed to extract signature decipher algorithm.` stdout message when creating the player, but this doesn't seem to work with the onesierequest approach. All content 403s.

absidue:
Okay so in conclusion this PR is working correctly, it's just that there are also issues with the nsig stuff that this PR wasn't intended to solve.

LuanRT (approved):

PR #844, opened 2024-12-14
feat(Playlists): Implement `addToLibrary` and `removeFromLibrary`

Duell10111:
Adds the possibility to add a external playlist to a user library.

Worked as sharm with v11, I adapted it to v12 structure. :)

### DIFF
--- a/src/core/managers/PlaylistManager.ts
+++ b/src/core/managers/PlaylistManager.ts
@@ -65,6 +65,44 @@ export default class PlaylistManager {
       data: response.data
     };
   }
+  
+  /**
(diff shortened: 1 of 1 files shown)


LuanRT (changes requested):
I think `addToLibrary` and `removeFromLibrary` would suit these methods better.

Duell10111:
> I think `addToLibrary` and `removeFromLibrary` would suit these methods better. Looks good otherwise.

Alright, renamed the methods. :)

Duell10111 (commented):
Method names changed.

LuanRT (approved):

PR #889, opened 2025-02-07
refactor: Remove internal uses of the `.first()` and `.matchCondition()` helpers

absidue:
As those `ObservedArray` helpers are just thin wrappers around `[0]` and `.find()` respectively. Using the original versions should be slightly faster and uses slightly less code, without losing any functionality.

It might be worth deprecating the `.first()`, `.matchCondition()` and `.remove()` helpers, because they all are just thin wrappers that don't provide any additional functionality than just interacting with the array yourself directly. Removing them straight away might break users code, hence the suggestion to deprecate them first.

### DIFF
--- a/src/core/Actions.ts
+++ b/src/core/Actions.ts
@@ -177,7 +177,7 @@ export default class Actions {
       let parsed_response = Parser.parseResponse<ParsedResponse<T>>(await response.json());
 
       // Handle redirects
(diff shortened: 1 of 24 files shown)


LuanRT:
Those are just helper functions for a better dev exp, so we won't be deprecating them (and sadly, I have some projects with thousands of lines that rely on these :p, it would be a pain to refactor that.)

I do agree that we shouldn't use them internally though.

LuanRT (approved):

PR #891, opened 2025-02-11
fix(Parser): parse all contents in home tab

ChunkyProgrammer:
This is a bit strange but it seems that the home pages on some channels(ex: https://www.youtube.com/channel/UC-9-kyTW8ZkZNDHQJ6FgpwQ ) have multiple different List/Grid renderers in the same object (not an array) so this PR allows for parsing them.

**To be decided/further questions**:
Is this the proper way to parse multiple items like this?
Should `content` get removed for `contents` since it could contain an empty sectionList (like the [music channel](https://www.youtube.com/channel/UC-9-kyTW8ZkZNDHQJ6FgpwQ) for example)? Or should we keep both properties (this wouldn't be a breaking change but might cause confusion because of potential duplication)

### DIFF
--- a/src/parser/classes/Tab.ts
+++ b/src/parser/classes/Tab.ts
@@ -12,6 +12,7 @@ export default class Tab extends YTNode {
   selected: boolean;
   endpoint: NavigationEndpoint;
   content: SectionList | MusicQueue | RichGrid | null;
+  contents: (SectionList | MusicQueue | RichGrid)[] | null;
 
   constructor(data: RawNode) {
     super();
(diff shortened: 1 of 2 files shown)


LuanRT:
Is there any channel where these extra renderers have any content? You provided the YouTube Music channel, but it only contains an empty renderer, as you noted. 

Also, I've noticed that on m.youtube.com, that section list is populated and  `disablePullToRefresh` is `true` (and this field is also present on that empty renderer on desktop, but it being there makes no sense), so perhaps this is a bug on YouTube's side?

ChunkyProgrammer:
I haven't found any other channels like this but i also wasn't looking for any.

ChunkyProgrammer:
I'll take a look at doing this a different way since we don't need to care about empty renderers

ghost:
I'm not really a fan of the code here but I'm not sure what should be done instead (the music channel is still empty it seems)

LuanRT:
Yeah this is bit of an edge case. I'll see if there's a better way.

PR #903, opened 2025-02-17
feat(parser): Add `AnimatedThumbnailOverlayView`

izakfilmalter:
Add AnimatedThumbnailOverlayView to ThumbnailView. Seem to run into it sporadically on some videos.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -74,6 +74,9 @@ deno/
 # VSCode files
 .vscode/
 
+# Webstorm files
+.idea/
+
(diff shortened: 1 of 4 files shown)


LuanRT (approved):

absidue:
Looks like this was merge prematurely. It has it's own image sources handling instead of using the `Thumbnail` class. @izakfilmalter Would you like to do a follow up PR to correct it or would you prefer someone else to do it?

PR #941, opened 2025-04-02
fix: replace import 'assert' with 'with' to compatible with node 22.

xiaohaoxing:


absidue:
Those two files are only used for React Native and in web browsers respectively, can you please explain why you think this pull request is necessary, considering that those files are not used for Node.js?

afkcodes:
hey this change will fix the error in react native 0.78+ @absidue this is the erro i am also getting in latest react native project

PR #951, opened 2025-04-22
feat(Session): Add option to override Player ID

paulrouget:
When YouTube API changes, sometimes just forcing `player_id` to an older value solves the issue. People have been doing this manually by overriding `const player_id = XXX` in the code itself.

Here I make it an option.


Nansess:
Hardcoding older player IDs increases the chances of your IP getting flagged, your account being flagged if you're using one, and receiving 403 errors. It raises the likelihood of all of these issues.

paulrouget:
When this library breaks, usually we need to wait a couple of days before it get fixed (which is amazingly fast!).

But I'm thinking of people who can't update the library as often as they want (for example, an android app needs to be repackaged, re-uploaded, reviewed, etc etc). Or in my case, I'm working on projects where remote JS code is not allowed and require external reviews. So my tools always lag a couple of weeks behind.

As a __temporary__ fix, using a Player ID works sometimes. But at the moment, it requires library update, which defeats the purpose (if I can update the library immediately, I wouldn't need such a workaround).

Nansess:
> PipePipe

completely diff PipePipe Runs on a local devices android to be exact so can use the android clients which don't require the JS player script

Nansess:
> When this library breaks, usually we need to wait a couple of days before it get fixed (which is amazingly fast!).
> 
> But I'm thinking of people who can't update the library as often as they want (for example, an android app needs to be repackaged, re-uploaded, reviewed, etc etc). Or in my case, I'm working on projects where remote JS code is not allowed and require external reviews. So my tools always lag a couple of weeks behind.
> 
> As a **temporary** fix, using a Player ID works sometimes. But at the moment, it requires library update, which defeats the purpose (if I can update the library immediately, I wouldn't need such a workaround).

Fair point. But keep in mind, these things change for a reason. Google/YouTube isn’t stupid—they don’t keep old scripts working for long. They also ramp up detection on outdated clients and scripts. Running old code is an easy way to get your IP, account, or both flagged, sadly when scraping the web you're dammed to deal with changes

src/core/Session.ts
```diff
@@ -205,6 +205,10 @@ export type SessionOptions = {
    * Proof of Origin Token. This is an attestation token generated by BotGuard/DroidGuard. It is used to confirm that the request is coming from a genuine client.
    */
   po_token?: string;
+  /**
+   * Player ID. Force a Player ID. Most time, this is not necessary. But when YouTube makes breaking API changes, using an older Player ID can help work around temporary breakage.
+   */
```
LuanRT, line 210:
```suggestion
  /**
   * Player ID override.
   * In most cases, this isn't necessary; but when YouTube introduces breaking changes,
   * forcing an older Player ID can help work around temporary issues.
   */
   ```

Issue #965, opened 2025-05-17
13.4.0 (Deno) -  Invalid JSON in Response


Jelenkee:
### Steps to reproduce

Run that script with Deno (2.3.3):

```js
import { Innertube } from "npm:youtubei.js";

const tube = await Innertube.create({
  retrieve_player: false,
});
console.log(await tube.search("test", { type: "video" }));
```

### Failure Logs

```shell
error: Uncaught (in promise) SyntaxError: Unterminated string in JSON at position 262029 (line 1 column 262030)
    at JSON.parse (<anonymous>)
    at Actions.execute (file:///home/gaus/.cache/deno/npm/registry.npmjs.org/youtubei.js/13.4.0/dist/src/core/Actions.js:91:24)
    at Object.runMicrotasks (ext:core/01_core.js:692:26)
    at processTicksAndRejections (ext:deno_node/_next_tick.ts:59:10)
    at runNextTicks (ext:deno_node/_next_tick.ts:76:3)
    at eventLoopTick (ext:core/01_core.js:185:21)
    at async Innertube.search (file:///home/gaus/.cache/deno/npm/registry.npmjs.org/youtubei.js/13.4.0/dist/src/Innertube.js:177:26)
    at async file:///home/gaus/Projects/buli/main.ts:64:13
```

### Expected behavior

I expect a search result without errors.

### Current behavior

There is an error when parsing the response.

### Version

Default

### Anything else?

The same error happens with `await tube.getChannel("UClCIWcZNvq15p0Y-E4ToGOw")`.

EDIT: Works as expected with NodeJS. So probably a Deno issue.

### Checklist

- [x] I am running the latest version.
- [x] I checked the documentation and found no answer.
- [x] I have searched the existing issues and made sure this is not a duplicate.
- [x] I have provided sufficient information.

Jelenkee:
I don't know if it's related, but when I add the global fetch to the config, I get an error with url serialization.

```js
const tube = await Innertube.create({
  retrieve_player: false,
  fetch: fetch
});
```

```
error: Uncaught (in promise) TypeError: Invalid URL: '[object Request]'
    at getSerialization (ext:deno_url/00_url.js:98:11)
    at new URL (ext:deno_url/00_url.js:405:27)
    at new Request (ext:deno_fetch/23_request.js:338:25)
    at ext:deno_fetch/26_fetch.js:374:29
    at new Promise (<anonymous>)
[...]

PR #967, opened 2025-05-21
refactor(Actions): Use Response.json instead of JSON.parse + Response.text

absidue:
May help with #965, even if it doesn't it is still a worthwhile clean up in my mind.

### DIFF
--- a/src/core/Actions.ts
+++ b/src/core/Actions.ts
@@ -179,7 +179,7 @@ export default class Actions {
     return {
       success: response.ok,
       status_code: response.status,
-      data: JSON.parse(await response.text())
+      data: await response.json()
     };
   }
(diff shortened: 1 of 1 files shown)


Jelenkee:
Unfortunately it doesn't help.
I added some console.log to the code and it seems that the body of the response is incomplete.
So using a different method for parsing JSON won't solve the problem.
I assume it is related to encoding (compression of the response)

absidue:
@Jelenkee Can you please check that deno is actually using the deno entrypoint in YouTube.js when installed from npm. One way to check is whether it works correctly if you install it from denoland which only has the deno entrypoint.

LuanRT (approved):

PR #974, opened 2025-06-08
feat(StreamingInfoOptions)!: Add `is_sabr` option

LuanRT:
Returns a manifest suitable for use in SABR player implementations.

NOTE: This changes the signature of `MediaInfo#toDash`, so I consider it a breaking change.

Example:
```ts
// ...
const info = await innertube.getBasicInfo('9Go0dVabZNs');
const manifest = await info.toDash({
  manifest_options: {
    is_sabr: true
  }
});
  
console.log(manifest);
```

### DIFF
--- a/src/core/mixins/MediaInfo.ts
+++ b/src/core/mixins/MediaInfo.ts
@@ -31,7 +31,7 @@ import type PlayerLiveStoryboardSpec from '../../parser/classes/PlayerLiveStoryb
 import type PlayerStoryboardSpec from '../../parser/classes/PlayerStoryboardSpec.js';
 
 export default class MediaInfo {
-  readonly #page: [IPlayerResponse, INextResponse?];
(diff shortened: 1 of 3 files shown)


src/core/mixins/MediaInfo.ts
```diff
@@ -113,28 +118,28 @@ export default class MediaInfo {
     let storyboards;
     let captions;
 
-    if (options.include_thumbnails && player_response.storyboards) {
+    if (manifest_options.include_thumbnails && player_response.storyboards) {
```
PikachuEXE, line 121: Should this use `include_thumbnails` from `options` or `manifest_options`?
Looking at `DashOptions` it seems to be the later but then the argument in `options` is misleading

LuanRT: The `options` argument type is wrong (will fix it later). Use `manifest_options`.

Issue #995, opened 2025-07-10
14.0.0 - `[YOUTUBEJS][Parser]: Error: DismissableDialogContentSection not found!`


feathecutie:
### Steps to reproduce

Minimal reproduction:

```ts
import { type IBrowseResponse, Innertube, Parser, YTNodes } from 'youtubei.js';

const innertube = await Innertube.create({});

const playlist = await innertube.music.getPlaylist('OLAK5uy_mvB1b-5JvguorHZ7EsoKx3jYDq4VDvt04');

const track = playlist.contents?.firstOfType(YTNodes.MusicResponsiveListItem);

const songCreditsEndpoint = track?.menu?.items.filterType(YTNodes.MenuNavigationItem).matchCondition((node) =>
	node.text === 'View song credits'
)
	?.endpoint;

if (songCreditsEndpoint) {
	const response = await songCreditsEndpoint.call(innertube.actions, { client: 'YTMUSIC' });
	Parser.parseResponse<IBrowseResponse>(response.data);
}
```

### Failure Logs

```shell
[YOUTUBEJS][Parser]: Error: DismissableDialogContentSection not found!
This is a bug, want to help us fix it? Follow the instructions at https://github.com/LuanRT/YouTube.js/blob/main/docs/updating-the-parser.md or report it at https://github.com/LuanRT/YouTube.js/issues!
Introspected and JIT generated this class in the meantime:
class DismissableDialogContentSection extends YTNode {
  static type = 'DismissableDialogContentSection';

  title: Text;
  subtitle: Text;

  constructor(data: RawNode) {
    super();
    this.title = new Text(data.title);
    this.subtitle = new Text(data.subtitle);
  }
}

    at ERROR_HANDLER (file:///home/fea/dev/projects/music/harmony/node_modules/.deno/youtubei.js@14.0.0/node_modules/youtubei.js/dist/src/parser/parser.js:69:27)
    at createRuntimeClass (file:///home/fea/dev/projects/music/harmony/node_modules/.deno/youtubei.js@14.0.0/node_modules/youtubei.js/dist/src/parser/generator.js:322:5)
    at generateRuntimeClass (file:///home/fea/dev/projects/music/harmony/node_modules/.deno/youtubei.js@14.0.0/node_modules/youtubei.js/dist/src/parser/generator.js:375:21)
[...]

PR #997, opened 2025-07-11
feat(Parser): Add `DismissableDialog` and `DismissableDialogContentSection` nodes

feathecutie:
Fixes https://github.com/LuanRT/YouTube.js/issues/995
Fixes https://github.com/LuanRT/YouTube.js/issues/996

This simply uses the suggested code from the JIT-generated classes, but from my testing, it seems to work great


LuanRT:
Thanks!

PR #1008, opened 2025-07-30
perf(Utils): Optimise `options.name` handling in `findVariable`

absidue:
This pull request hoists the `options.name` check outside of the for-loop, that way if the option is specified we only have to check it once and when it is not specified we can skip the for-loop entirely.


absidue:
Not needed anymore

PR #1020, opened 2025-08-19
Decipher url using video bound po_token

iBicha:
This is kind of a follow up to https://github.com/LuanRT/YouTube.js/pull/994

Since `PoToken` can be videoId bound, then we should use video specific tokens for decoding (which is just setting the `pot` param`) instead of using the one in the Player, which I assume would be the session token.

I started making the changes, but then I didn't know which direction I wanted to go, because there are more changes that needed to be done, and I found myself passing po_token arg everywhere.

I thought I'd draft this and get some feedback first, or at least bring attention to it

### DIFF
--- a/src/Innertube.ts
+++ b/src/Innertube.ts
@@ -463,7 +463,7 @@ export default class Innertube {
     const info = await this.getBasicInfo(video_id, options);
 
     const format = info.chooseFormat(options);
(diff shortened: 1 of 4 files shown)


iBicha:
Since only /player calls seem to need a videoId bound poToken, but not the adaptive format urls themselves, perhaps this is not needed

PR #1042, opened 2025-09-18
feat(getTrending): Add support for searching different trending pages

Fijxu:
Add the ability to search different trending pages instead of the default one.

Acording to https://github.com/iv-org/invidious/issues/5397#issuecomment-3218928458 and https://github.com/TeamNewPipe/NewPipe/discussions/12445, Youtube is removing their aggregated trending page. This adds a few more trending options instead of always fetching the default trending page.

### DIFF
--- a/src/Innertube.ts
+++ b/src/Innertube.ts
@@ -52,6 +52,13 @@ import {
   SearchFilter_SortBy
 } from '../protos/generated/misc/params.js';
 
+export enum TrendingPage {
+  Gaming = 0,
+  Movies = 1,
+  Music = 2,
+  Livestreams = 3,
+}
+
 /**
  * Provides access to various services and modules in the YouTube API.
  *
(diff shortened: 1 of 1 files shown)


src/Innertube.ts
```diff
@@ -52,6 +52,13 @@ import {
   SearchFilter_SortBy
 } from '../protos/generated/misc/params.js';
 
+export enum TrendingPage {
+  Gaming = 0,
+  Movies = 1,
+  Music = 2,
+  Livestreams = 3,
+}
+
```
Fijxu, line 61: This should be moved to another place, but I'm unsure where to put it. Please tell me where to move it if necessary ;)

src/Innertube.ts
```diff
@@ -364,8 +371,27 @@ export default class Innertube {
...
+    switch (trendingPage) {
+      case TrendingPage.Gaming:
+        params = '4gIcGhpnYW1pbmdfY29ycHVzX21vc3RfcG9wdWxhcg%3D%3D';
+        break;
+      case TrendingPage.Movies:
+        params = '4gIKGgh0cmFpbGVycw%3D%3D';
+        break;
+      case TrendingPage.Music:
+        params = '4gINGgt5dG1hX2NoYXJ0cw%3D%3D';
+        break;
```
Fijxu, line 387: Taken from https://github.com/iv-org/invidious/blob/325e013e0d9e5670fa0df7635ff30a0ee029e05e/src/invidious/trending.cr#L8-L13. I'm unsure if they are getting to be deprecated by Youtube

Fijxu:
Closed since `(await innertubeClient.getTrending()).tabs` and `await (await innertubeClient.getTrending()).getTabByName(tabName)` can be used.

Reference: https://github.com/iv-org/invidious-companion/issues/201#issuecomment-3311211712

PR #1075, opened 2025-10-18
fix(history): Add support for `LockupView` content

dnicolson:
This adds support for `LockupView` content in the watch history in addition to `Video` content. It seems that recently YouTube started using `LockupView` content in almost every instance.

### DIFF
--- a/src/parser/youtube/History.ts
+++ b/src/parser/youtube/History.ts
@@ -5,8 +5,8 @@ import Button from '../classes/Button.js';
 
 import type { Actions, ApiResponse } from '../../core/index.js';
 import type { IBrowseResponse } from '../types/index.js';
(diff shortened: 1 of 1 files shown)


dnicolson:
This code has worked reliably since the PR was created. In one history continuation, just 6 of 200 history items were of type `Video`, while the remainder were `LockupView`.

absidue (commented):
test (GitHub won't let me leave any line specific review comments).

absidue (commented):
As GitHub won't let me comment on specific lines on this pull request, I'll leave a comment like this instead.

Please use the `YTNode.is()` type guard instead of explicitly checking the type property and casting.

dnicolson:
@absidue is it possible to reopen this please? It is a very useful fix.

PR #1077, opened 2025-10-19
Check if duration exists

dnicolson:
This prevents the following for live and upcoming videos:
```
{text: 'N/A', seconds: NaN}
```

### DIFF
--- a/src/parser/classes/PlaylistVideo.ts
+++ b/src/parser/classes/PlaylistVideo.ts
@@ -25,7 +25,7 @@ export default class PlaylistVideo extends YTNode {
   accessibility_label?: string;
   style?: string;
 
-  duration: {
+  duration?: {
     text: string;
     seconds: number;
   };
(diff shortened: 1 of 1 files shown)


absidue:
Please fix the type error.

dnicolson:
The issue can be reproduced by adding a live video to the Watch Later playlist:
```
const VIDEO_ID = "jfKfPfyJRdk";
const PLAYLIST_ID = "WL";

await yt.playlist.addVideos(PLAYLIST_ID, [VIDEO_ID]);
const playlist = await yt.getPlaylist(PLAYLIST_ID);

for (const item of playlist.items) {
  if (item.type === "PlaylistVideo" && item.id === VIDEO_ID) {
    console.log(`Duration of video ${VIDEO_ID}`, item.duration);
    break;
  }
}
```

dnicolson:
@LuanRT this might still be useful.

Issue #1038, opened 2025-09-16
[15.1.1] React native export entry point


lovegaoshi:
### Question

Hi! 

I recently upgraded my react native project to use esm import only and started to see `super expression must either be null or a function`. Then I noticed ytbi used to be imported via the bundled src `./bundle/react-native.js`, however its done that way, to now via `./dist/src/platform/react-native.js`. Replacing `"react-native": "./dist/src/platform/react-native.js",` to `"react-native": "./bundle/react-native.js",` fixes it. I wonder what's the decision on all entry points being at ./dist/src instead of ./bundle?

### Checklist

- [x] I am running the latest version.
- [x] I checked the documentation and found no answer.
- [x] I have searched the existing issues and made sure this is not a duplicate.
- [x] I have provided sufficient information.

LuanRT:
To answer this, I'd like to get rid of all the single file bundles eventually. Did you find a way around this or does it only work if you modify that?

absidue:
The only differences between the standard export and the bundle are:
- Everything is in one file
- It has polyfills for modern JavaScript syntax

Please check that you are using the latest version of React Native, have babel setup correctly in your project and all other necessary polyfills.

lovegaoshi:
found the culprit is `node_modules\youtubei.js\dist\src\parser\classes\Video.js` cannot be imported by VideoCard.js for whatever reason. if i duplicate Video.js to another file and import from there everything works.

i remember this was an issue some times ago though.

PR #1094, opened 2025-11-16
Add additional browser target

dnicolson:
This adds a build that is usable in the browser without any transpilation.

Presently, you would need to run this to use the project in a script tag:
```
npm install youtubei.js
npx webpack ./node_modules/youtubei.js/bundle/browser.js --output-library YouTube --output-library-type umd -o youtubei.umd.js
```

With this change, it can be used from a Userscript:
```
// @require https://cdn.jsdelivr.net/npm/youtubei.js@latest/bundle/browser-iife.js

// ==UserScript==
    const Innertube = YouTubeJS.default
    const yt = await Innertube.create({
        cookie: document.cookie,
        fetch: (...args) => fetch(...args)
    });
// ==
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -83,6 +83,7 @@
     "build:proto": "rimraf ./protos/generated && node ./dev-scripts/generate-proto.mjs",
     "build:parser-map": "node ./dev-scripts/gen-parser-map.mjs",
     "bundle:browser": "esbuild ./dist/src/platform/web.js --banner:js=\"/* eslint-disable */\" --bundle --sourcemap --target=chrome70 --keep-names --format=esm --define:global=globalThis --conditions=module --outfile=./bundle/browser.js --platform=browser",
+    "bundle:browser-iife": "esbuild ./dist/src/platform/web.js --bundle --sourcemap --target=es2017 --keep-names --format=iife --global-name=YouTubeJS --outfile=./bundle/browser-iife.js --platform=browser",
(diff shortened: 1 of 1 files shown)


absidue:
As mentioned in this comment https://github.com/LuanRT/YouTube.js/issues/1038#issuecomment-3394301801 we would like to remove the prebuilt bundles, not add more. If you have any special requirements like it requiring it to be wrapped in an IIFE, you should handle that on your side.

dnicolson:
The only way to handle it is to bundle the library and host it somewhere whenever a new version is released. The chrome.userScripts API doesn't support ESM, YouTube prevents dynamic imports via CSP, and there doesn't seem to be a CDN like Skypack that can deliver an IIFE build.

In case it helps anyone, this repo automatically generates UMD builds:
https://dnicolson.github.io/youtubei.js-umd/latest/youtubei.js.umd.js

absidue:
@dnicolson It would probably make sense to run your user script through a bundler, it would make it smaller and faster than shipping a UMD bundle of YouTube.js. YouTube.js is meant to be used as a library with a bundler, it uses a lot of namespaces which a bundler can convert to direct function and class references but in a UMD build all of those namespaces become objects, which not only means more code but also extra overhead at runtime, because everything requires property accesses.

dnicolson:
@absidue I’ve discovered over the years that bundling Userscripts is generally a bad idea. Most browser extensions hang due to parsing, syntax highlighting, etc., and it makes debugging more cumbersome.

PR #1097, opened 2025-12-02
fix: respect language parameter when filtering by itag in chooseFormat

vlrdimir:
When multiple formats share the same itag but differ in language (e.g., original vs dubbed audio tracks), `chooseFormat` now correctly filters by the `language` parameter instead of always returning the first match.

## Changes

- Filter by `is_original` flag when `language = 'original'`
- Filter by exact language code for specific languages (e.g., `'en-US'`)
- Fallback to first format when requested language is not available
- Add warning log when language not found (shows available languages)
- Set default `language = 'original'` to match documented behavior
- Fix audio selection for videos with multiple language tracks

## Before this fix

- Videos with multiple audio tracks (original + dubbed) would always
  download the first format with matching itag (usually German dubbed)
- `language` parameter was ignored when `itag` was specified
- No warning when requested language was unavailable

## After this fix

- Correctly selects original audio when `language = 'original'`
- Respects specific language codes when provided
- Warns user when requested language is not found
- Maintains backward compatibility with graceful fallback

## Example

Video: https://www.youtube.com/watch?v=M_W-dleZXCs  
Has 15 audio tracks (12 dubbed + 3 original) all with `itag = 140`:

- **Before:** Selected German (DE) dubbed audio (first in list)
- **After:** Correctly selects English (US) original audio

### DIFF
--- a/src/utils/FormatUtils.ts
+++ b/src/utils/FormatUtils.ts
@@ -1,5 +1,6 @@
 import * as Constants from './Constants.js';
 import { InnertubeError, Platform, streamToIterable } from './Utils.js';
+import * as Log from './Log.js';
 
 import type Player from '../core/Player.js';
 import type Actions from '../core/Actions.js';
@@ -26,6 +27,7 @@ export async function download(
     quality: '360p',
     type: 'video+audio',
     format: 'mp4',
+    language: 'original',
     range: undefined,
     ...options
   };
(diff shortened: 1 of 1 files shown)


iBicha:
Opinion; I would solve this slightly differently
Instead of selecting best candidate, then falling back to a random option, I would retain the same pattern to be strict with format selection, and throwing an error if it's not available.

Then, I would change the API to accept either a single `options: FormatOptions`, or an array of options objects.
That way, the caller can specify themselves fallback options, loosening their selection criteria in the array.
`chooseFormat` would loop through the options array, until it finds a matching format. It would still throw an error if no matches are found within all options objects.

PR #1104, opened 2025-12-26
feat(core): Add getCaptionTrackContent (MediaInfo)

### LINKED ISSUE #1102: 16.0.1 Error: Request to https://www.youtube.com/youtubei/v1/get_transcript?prettyPrint=false&alt=json failed with status 400
### Steps to reproduce

1. Call `info.getTranscript();` in nodejs server

```js

try {
    const yt = await Innertube.create({
      generate_session_locally: true,
      lang: language,
      cookie: COOKIE
    });

    const info = await yt.getInfo(video_url);

    title = info.primary_info?.title ? info.primary_info?.title.toString() : '';

    const defaultTranscriptInfo = await info.getTranscript();

console.log(defaultTranscriptInfo)

} catch(e) {
console.log(e)
}
```

### Failure Logs

```shell
Error: Request to https://www.youtube.com/youtubei/v1/get_transcript?prettyPrint=false&alt=json failed with status 400
```

### Expected behavior

it supposed to return the transcript 

### Current behavior

It returns that error since last week , sometimes it works after a couple of tries like 10+ tries

### Version

Default

### Anything else?

_No response_

### Checklist

- [x] I am running the latest version.
- [x] I checked the documentation and found no answer.
- [x] I have searched the existing issues and made sure this is not a duplicate.
- [x] I have provided sufficient information.

limdingwen:
Somewhat fixes #1102

Adds an alternate API for getting captions by using the caption track's base URL. Motivation for this is fixing Invidious Companion, and because yt-dlp also uses the caption API for getting the transcript.

This is an alternate API because it seems like captions return different things from the searcheable transcript? I'm not too sure, though.

Tests are not in yet. WIP.

Example code:
```
import { Innertube } from "./platform/node.js";

(async () => {
  const innertube = await Innertube.create({
    generate_session_locally: true,
    lang: "en",
    location: "US",
  });
  const info = await innertube.getInfo("k7fXbdRH9v4");
  const captions = info.captions;
  if (!captions) {
    console.log("Captions is undefined");
    return;
  }
  const captionTracks = captions.caption_tracks;
  if (!captionTracks) {
    console.log("Caption tracks is undefined");
    return;
  }
  const captionTrack = captionTracks[0];
  const response = await info.getCaptionTrackContent(captionTrack);
  console.log(response);
})();
```

### DIFF
--- a/src/core/mixins/MediaInfo.ts
+++ b/src/core/mixins/MediaInfo.ts
@@ -27,6 +27,7 @@ import type CardCollection from '../../parser/classes/CardCollection.js';
 import type Endscreen from '../../parser/classes/Endscreen.js';
 import type PlayerAnnotationsExpanded from '../../parser/classes/PlayerAnnotationsExpanded.js';
 import type PlayerCaptionsTracklist from '../../parser/classes/PlayerCaptionsTracklist.js';
+import type { CaptionTrackContent, CaptionTrackContentLine, CaptionTrackData } from '../../parser/classes/PlayerCaptionsTracklist.js';
 import type PlayerLiveStoryboardSpec from '../../parser/classes/PlayerLiveStoryboardSpec.js';
 import type PlayerStoryboardSpec from '../../parser/classes/PlayerStoryboardSpec.js';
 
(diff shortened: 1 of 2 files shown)


limdingwen:
For those coming from Invidious Companion, here's the code needed to make it work with this new API. I haven't opened a PR there though, since I don't want to spam PRs.

```
import { Innertube } from "youtubei.js";
import type { CaptionTrackData } from "youtubei.js/PlayerCaptionsTracklist";
import { HTTPException } from "hono/http-exception";

function createTemporalDuration(milliseconds: number) {
    return new Temporal.Duration(
        undefined,
        undefined,
        undefined,
        undefined,
        undefined,
        undefined,
        undefined,
        milliseconds,
    );
}

const ESCAPE_SUBSTITUTIONS = {
    "&": "&amp;",
    "<": "&lt;",
    ">": "&gt;",
    "\u200E": "&lrm;",
    "\u200F": "&rlm;",
    "\u00A0": "&nbsp;",
};

export async function handleTranscripts(
    innertubeClient: Innertube,
    videoId: string,
    selectedCaption: CaptionTrackData,
) {
    const lines: string[] = ["WEBVTT"];

    const info = await innertubeClient.getInfo(videoId);
    const captions = info.captions;
    if (captions == undefined) throw new HTTPException(404);
    const captionTracks = captions.caption_tracks;
    if (captionTracks == undefined) throw new HTTPException(404);
    console.log(captionTracks);

    const captionTrack = captionTracks.find((
        ct,
    ) => {
        const inner = selectedCaption.name.text;
        const outer = ct.name.text;
        if (inner == undefined) return false;
        if (outer == undefined) return false;
        return outer.toLowerCase().includes(inner.toLowerCase());
    });
    if (captionTrack == undefined) throw new HTTPException(404);
    const captionTrackContent = await info.getCaptionTrackContent(captionTrack);

    captionTrackContent.lines.forEach((line) => {
        const timestampFormatOptions = {
            style: "digital",
            minutesDisplay: "always",
            fractionalDigits: 3,
        };

        // Temporal.Duration.prototype.toLocaleString() is supposed to delegate to Intl.DurationFormat
        // which Deno does not support. However, instead of following specs and having toLocaleString return
        // the same toString() it seems to have its own implementation of Intl.DurationFormat,
        // with its options parameter type incorrectly restricted to the same as the one for Intl.DateTimeFormatOptions
        // even though they do not share the same arguments.
        //
        // The above matches the options parameter of Intl.DurationFormat, and the resulting output is as expected.
        // Until this is fixed typechecking must be disabled for the two use cases below
        //
        // See
        // https://docs.deno.com/api/web/~/Intl.DateTimeFormatOptions
        // https://docs.deno.com/api/web/~/Temporal.Duration.prototype.toLocaleString
        // https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Temporal/Duration/toLocaleString
        // https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/DurationFormat/DurationFormat

        const start_ms = createTemporalDuration(
            Number(Math.floor(line.start * 1000)),
        ).round({
            largestUnit: "year",
            relativeTo: Temporal.PlainDateTime.from("2022-01-01"),
            //@ts-ignore see above
        }).toLocaleString("en-US", timestampFormatOptions);

        const end_ms = createTemporalDuration(
            Number(Math.floor(line.end * 1000)),
        ).round({
            largestUnit: "year",
            relativeTo: Temporal.PlainDateTime.from("2022-01-01"),
            //@ts-ignore see above
        }).toLocaleString("en-US", timestampFormatOptions);
        const timestamp = `${start_ms} --> ${end_ms}`;

        const text = (line.text || "").replace(
            /[&<>‍‍\u200E\u200F\u00A0]/g,
            (match: string) =>
                ESCAPE_SUBSTITUTIONS[
                    match as keyof typeof ESCAPE_SUBSTITUTIONS
                ],
        );

        lines.push(`${timestamp}\n${text}`);
    });

    return lines.join("\n\n");
}
```

absidue:
If you want to parse the captions you should do that in your own project, you definitely shouldn't be "parsing" XML with basic find and replace operations.

PR #1120, opened 2026-01-20
feat(Parser): Add ThumbnailOverlayTitleView and PageIndicatorView classes

zestysoft:
## Summary

Add two missing parser classes that were being JIT-generated at runtime:

- **ThumbnailOverlayTitleView**: Displays title/subtitle overlays on thumbnails (e.g., "LIVE", "MIX" labels)
- **PageIndicatorView**: Tracks pagination state in carousels (indicator count and selected index)

These are encountered when parsing Mix/Playlist responses and were causing console warnings:

```
[YOUTUBEJS][Parser]: Error: ThumbnailOverlayTitleView not found!
[YOUTUBEJS][Parser]: Error: PageIndicatorView not found!
```

## Test plan

- [x] Build completes successfully with `npm run build`
- [x] Parser map auto-generates exports in `nodes.ts`

### DIFF
--- a/src/parser/classes/PageIndicatorView.ts
+++ b/src/parser/classes/PageIndicatorView.ts
@@ -1,0 +1,15 @@
+import { YTNode } from '../helpers.js';
+import type { RawNode } from '../index.js';
+
+export default class PageIndicatorView extends YTNode {
+  static type = 'PageIndicatorView';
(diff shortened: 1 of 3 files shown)


LuanRT:
Thanks!

PR #1163, opened 2026-05-03
fix(Channel): Add support for new feed filter nodes

LuanRT:
See title. I was supposed to fix this a long time ago, but things got busy again. It is what it is, I suppose.

Example usage:
```ts
// ...

const channel = await innertube.getChannel('UCy8E4XShrcph5WWLg05zETQ');
const videos = await channel.getVideos();
  
// Apply a primary filter by name.
const filtered = await videos.applyFilter('Oldest');

// Apply a primary and secondary filter by name.
const filtered = await videos.applyFilter('Oldest', 'Members only');

// Since we're using `filtered`, the following will return the latest members-only videos, 
// unless the secondary filter is explicitly changed.
const latestMembersOnly = await filtered.applyFilter('Latest');
```


LuanRT:
No idea what's up with the YouTube Music lyrics test case, but I'll look into that later.

absidue (approved):

PR #1174, opened 2026-05-14
fix(ThumbnailBadgeView): Parse icon

gBasil:
This PR fixes the setting of the `ThumbnailBadgeView.icon_name` field. The old check for `data.iconName` no longer succeeds, even on badges that have icons, presumably because the field was removed at some point, so I changed it to check for `data.icon` instead.

I encountered this on the home feed (`getHomeFeed()`) for music and livestreams, so that's one avenue for testing it.


LuanRT (approved):

LuanRT:
Good catch.

PR #1203, opened 2026-07-12
feat(Collaborators): Add `Author#collaborators` and `Playlist#getCollaborators()`

Shadorc:
- Add `get collaborators()` inside `Author` class to add support for collaborators in search page data.
- Add `getCollaborators()` inside `Playlist` class to add support for colaborative playlists ([example](https://www.youtube.com/playlist?list=PLaen7aJT3lIFptMwtTJkOvQtcxyKqQwlP)), requires authentication

This does not work in channel ([example](https://www.youtube.com/@KSI/videos)) or playlist ([example](https://www.youtube.com/playlist?list=PL6_iWvoCGAJlJUy9KFfGX9e41ZHB37wGX)) because YouTube does not provide this data.
Video page already supports getting collaborators through `LockupView` fields.

Playlists themselves can also be collaborative ([example](https://www.youtube.com/playlist?list=PLaen7aJT3lIFptMwtTJkOvQtcxyKqQwlP)), but it requires an extra API call and it is not accessible without being authentified.

### Code Sample
```javascript
const innertube = await Innertube.create()
const searchResults = await innertube.search('KSI')
const videos = searchResults.results.filter(result => result.type === 'Video')
const collaborators = videos.map(video => video.author?.collaborators
    .map(collaborator => (
        {
            id: collaborator.renderer_context?.command_context?.on_tap?.payload?.browseId,
            name: collaborator.title?.text,
            thumbnails: collaborator.leading_accessory,
            url: collaborator.renderer_context?.command_context?.on_tap?.toURL(),
            channel_info: collaborator.subtitle?.text,
        }
    )))


console.log(JSON.stringify(collaborators, null, 2))
```

```javascript
const innertube = await Innertube.create({ cookie: "" })
const playlistInfo = await innertube.getPlaylist('PLaen7aJT3lIFptMwtTJkOvQtcxyKqQwlP')

const collaborators = await playlistInfo.getCollaborators()
console.log(JSON.stringify(collaborators, null, 2))
```


LuanRT:
I think this should probably be a getter instead. Something like this:
```ts
// ...
get collaborators(): ListItemView[] {
  if (this.endpoint?.command?.is(ShowDialogCommand) && this.endpoint.command.inline_content?.is(DialogView)) {
    const dialog = this.endpoint.command.inline_content;

    if (dialog.custom_content?.is(ListView)) {
      return dialog.custom_content.items.as(ListItemView)
        .filter((item) => item.renderer_context?.command_context?.on_tap?.metadata?.page_type === 'WEB_PAGE_TYPE_CHANNEL');
    }
  }

  return [];
}
```

I also recommend removing that `Collaborator` class. Itt seems redundant and quite sparse compared to the original `ListItemView` node. And if you need info like channel urls, `someCollaborator.renderer_context?.command_context?.on_tap?.toURL()` should work.

LuanRT:
> Playlists themselves can also be collaborative, but it does not seem to be supported by YouTube's API. The response from Innertube (`Innertube#getPlaylist`) does not contain any mention to collaborators other than their thumbnails. Example: https://www.youtube.com/playlist?list=PLaen7aJT3lIFptMwtTJkOvQtcxyKqQwlP

Looks like it requires an extra API call. In that case, you could implement a `getCollaborators()` method, provided the right nav endpoint is available in the response.

Shadorc:
Thanks for the suggestions!

Edit: I was able to make the `get_panel` endpoint works, so I removed all my questions

Shadorc:
I was finally able to figure out how to create the correct request to add support for `getCollaborators()` inside `Playlist`.
MR updated!

protos/misc/params.proto
```diff
@@ -266,3 +266,11 @@ message CommunityPostCommentsParam {
   string title = 2;
   CommentDataContainer comment_data_container = 53;
 }
+
+message PlaylistCollaboratorParams {
```
LuanRT: If fetching collaborators is only supported when logged in, maybe we should just take the params from the API response? Could probably result in something simpler.

src/parser/classes/endpoints/ShowEngagementPanelEndpoint.ts
```diff
@@ -1,15 +1,36 @@
...
+  buildRequest(): GetPlaylistCollaboratorsRequest {
+    const request: GetPlaylistCollaboratorsRequest = {};
+
+    if (this.#data.panelIdentifier) 
+      request.panelId = this.#data.panelIdentifier;
+
+    if (this.#data.sourcePanelIdentifier)
+      request.params = this.#data.sourcePanelIdentifier;
+
+    return request;
```
LuanRT: By the way, the endpoint classes are based on code from polymer (ex: https://www.youtube.com/s/desktop/4965577f/jsbin/desktop_polymer.vflset/desktop_polymer.js).

This is what `ShowEngagementPanelEndpoint` looks like from a quick look:
```js
    sOa = ["get_panel"]
    var WKa = new C("showEngagementPanelEndpoint");
    var sA = function() {};
    m(sA, Sy);
    sA.prototype.getApiPaths = function() {
        return sOa
    }
    ;
    sA.prototype.getExtension = function(a) {
        return E(a, WKa)
    }
    ;
    sA.prototype.buildRequest = function(a, b) {
        var c, d = b.panelIdentifier || (null == (c = b.identifier) ? void 0 : c.tag);
        d && (a.panelId = d);
        var f;
        if (null == (f = b.globalConfiguration) ? 0 : f.params) {
            var h;
            a.params = null == (h = b.globalConfiguration) ? void 0 : h.params
        }
    }
    ;
```

src/parser/youtube/Playlist.ts
```diff
@@ -68,6 +70,27 @@ export default class Playlist extends Feed<IBrowseResponse> {
...
+
+    const params = encodeURIComponent(u8ToBase64(writer.finish()).replace(/\+/g, '-').replace(/\//g, '_'));
+    const get_collaborators_endpoint = new NavigationEndpoint({
+      showEngagementPanelEndpoint: {
+        panelIdentifier: 'PAplaylist_collaborate',
+        sourcePanelIdentifier: params
+      }
+    });
+
+    return get_collaborators_endpoint.call(this.actions);
```
LuanRT: We should parse this. I can help with it later if you're unsure how to go about it.

protos/misc/params.proto
Shadorc: Indeed, it would be simpler.
The `params` is here from the get playlist response: `response.data.header.pageHeaderRenderer.content.pageHeaderViewModel.metadata.contentMetadataViewModel.metadataRows[0].metadataParts[0].avatarStack.avatarStackViewModel.rendererContext.commandContext.onTap.innertubeCommand.showEngagementPanelEndpoint.globalConfiguration.params` 

Now, I need to get it properly. Maybe some of this path is already parsed somewhere, but I probably have to implement some of them. 

**Update 1:**
I was able to correctly get the params from the API response, but it raises one question.
Doing a get_panel request on a playlist with a single collaborator (what most playlists are) gives a [valid response](https://gist.github.com/Shadorc/ecb7f18fb6b9074674b0b619caed8f7d).
However, `params` is not part of the response for such playlist. It is then not possible to do the request, even if that has no real purpose.

I think it will be solved when parsing the `getCollaborators` API response. It would then be possible to return some default response with no interesting data, like YouTube does.

**Update 2:**
I was wondering,
> If fetching collaborators is only supported when logged in

this approach could have been done even if the endpoint was supported when logged out, no?

src/parser/classes/endpoints/ShowEngagementPanelEndpoint.ts
Shadorc: Interesting!
So, it should look like this?

```typescript
buildRequest(): GetPlaylistCollaboratorsRequest {
  const request: GetPlaylistCollaboratorsRequest = {};

  const panelId = this.#data.panelIdentifier || this.#data.identifier?.tag;
  if (panelId) 
    request.panelId = panelId;

  if (this.#data.globalConfiguration?.params)
    request.params = this.#data.globalConfiguration?.params;

  return request;
}
```

**Update 1:**
`GetPlaylistCollaboratorsRequest` should probably be renamed to `ShowEngagementPanelRequest` too

**Update 2:**
Done!

Shadorc: `GetPlaylistCollaboratorsRequest` should probably be renamed to `ShowEngagementPanelRequest` too

src/parser/youtube/Playlist.ts
```diff
@@ -73,13 +73,12 @@ export default class Playlist extends Feed<IBrowseResponse> {
     if (!this.actions.session.logged_in)
       throw new Error('You must be signed in to perform this operation.');
 
-    const writer = PlaylistCollaboratorParams.encode({
-      params: {
-        playlistId: this.endpoint?.payload.playlistId
-      }
-    });
+    const avatarStackView = this.memo.getType(AvatarStackView)?.find((item) => item.renderer_context.command_context);
+    const params = avatarStackView?.renderer_context.command_context?.on_tap?.payload?.globalConfiguration?.params;
```
LuanRT: `on_tap` is a nav endpoint, and since you fixed its implementation (`ShowEngagementPanelEndpoint` class), you can call it directly (no need to build the request manually like it's done below this).

src/parser/youtube/Playlist.ts
Shadorc: I will have to take a look, not really sure for now how to achieve that.

src/parser/youtube/Playlist.ts
Shadorc: Indeed, thanks!

src/parser/youtube/Playlist.ts
Shadorc: I've added [a very basic parsed class](https://github.com/Shadorc/YouTube.js/blob/21baebc849d1a0aca335605ff44458e4de4da25d/src/parser/youtube/PlaylistCollaborate.ts), just enough to understand how it should work. 
Let me know what kind of changes I should make, or feel free to make them yourself if it's easier for you

LuanRT: Nice!
I don't think we need a whole class for it, just something in the `Playlist` class itself would work.
What we don't want is raw InnerTube responses being returned. For example, your class looks like this when logged to the console, which is inconsistent with the rest of the library:
```ts
PlaylistCollaborate {
  content: {
    engagementPanelSectionListRenderer: {
      header: [Object],
      content: [Object],
      veType: 211189,
      targetId: 'PAplaylist_collaborate',
      identifier: [Object]
    }
  }
}
```

protos/misc/params.proto
LuanRT: > **Update 2:** I was wondering,
> 
> > If fetching collaborators is only supported when logged in
> 
> this approach could have been done even if the endpoint was supported when logged out, no?

Oh sorry, yes.

LuanRT:
Probably ready to merge.

Small example for the updated `Playlist#getCollaborators` code:
```ts
import { Innertube, UniversalCache, YTNodes } from 'youtubei.js';

const innertube = await Innertube.create({ cache: new UniversalCache(true), cookie: "your_cookie_here" });

const playlist = await innertube.getPlaylist('PLaen7aJT3lIFptMwtTJkOvQtcxyKqQwlP');
const collab = await playlist.getCollaborators();

const sectionList = collab.content?.as(YTNodes.EngagementPanelSectionList);
const collaborationView = sectionList?.content?.as(YTNodes.PlaylistCollaborationView);
console.log(collaborationView?.playlist_collaborators);
```

Shadorc:
Very interesting, thank you for your help! 
I was wondering, how do you determine all the fields and values required to create the YTnode parsed? 
Like those enums: https://github.com/LuanRT/YouTube.js/pull/1203/changes/20bd8da218f0ef50420761a10cabf366994e62f0#diff-58d0d0e1e3a6a52f4482d42c001ab5f1da64edab62a0965482eb608a0504ed07R18

LuanRT:
> Very interesting, thank you for your help! I was wondering, how do you determine all the fields and values required to create the YTnode parsed? Like those enums: [20bd8da#diff-58d0d0e1e3a6a52f4482d42c001ab5f1da64edab62a0965482eb608a0504ed07R18](https://github.com/LuanRT/YouTube.js/commit/20bd8da218f0ef50420761a10cabf366994e62f0#diff-58d0d0e1e3a6a52f4482d42c001ab5f1da64edab62a0965482eb608a0504ed07R18)

I use this as a ref: https://drive.proton.me/urls/X5VYN77CTW#4LGdKdXaAINf (or https://github.com/LuanRT/youtubei-dts, as it makes things easier to understand).

And when that isn't available, I look at the native apps, since they sometimes don't obfuscate the protobuf messages (the iOS app is the most legible one).

LuanRT (approved):

Issue #1164, opened 2026-05-05
This will fail with 400 at setVideoMetadata


OurLobanov:
### Describe the bug

When attempting to upload a video using `yt.studio.upload()`, the process fails at the final `#setVideoMetadata` step with a `400 Bad Request` (INVALID_ARGUMENT) error. 

The video chunk upload to Scotty (`#uploadVideo`) completes successfully and returns a valid `scottyResourceId` with `STATUS_SUCCESS`. However, the subsequent request to `/upload/createvideo` is rejected by YouTube.

After intercepting the actual browser request for `/upload/createvideo` from YouTube Studio Web, it appears Google has updated the required payload and introduced a BotGuard (`po_token`) check for this endpoint.

### Reproduction Steps

```javascript
import { Innertube, UniversalCache } from "youtubei.js";
import { readFileSync } from "fs";

(async () => {
  // Using valid cookies
  const yt = await Innertube.create({ cookie: "SAPISID=...; SID=...", cache: new UniversalCache(false) });
  
  const file = readFileSync("./test.mp4");
  
  try {
    // This will fail with 400 at setVideoMetadata
    const uploadResult = await yt.studio.upload(file, {
      title: "Test Upload",
      description: "Test description",
      privacy: "UNLISTED"
    });
    console.log(uploadResult);
  } catch (error) {
    console.error("Upload failed:", error);
  }
})();
```

### Expected Behavior

The video should be successfully published/created, and `yt.studio.upload()` should return the new video metadata.

### Actual Behavior

Throws an `InnertubeError`:
```json
{
  "error": {
    "code": 400,
    "message": "Request contains an invalid argument.",
    "errors": [
      {
        "message": "Request contains an invalid argument.",
        "domain": "global",
        "reason": "badRequest"
      }
    ],
    "status": "INVALID_ARGUMENT"
  }
}
```

### Additional Context / Proposed Fix

I intercepted the `fetch` request from `studio.youtube.com` during a successful manual upload. There are several significant differences between `youtubei.js`'s current `Studio.js` payload and the real web payload:

1. **BotGuard (`attestationResponseData`) is now required**: 
The browser sends a `context.request.attestationResponseData` object with a `challenge` and a `webResponse` (starting with `$gB85...`). Without this Proof of Origin (`po_token`), YouTube rejects the request.
```json
"context": {
  "request": {
    "attestationResponseData": {
      "challenge": "...",
[...]

Issue #418, opened 2023-06-20
Add Upload Progress Feature


3IMAD69:
### Describe your suggestion

I would like to suggest adding an upload progress feature . Currently, there is no built-in functionality to track the progress of file uploads, which can be helpful.

### Checklist

- [X] I am running the latest version.
- [X] I checked the documentation and found no answer.
- [X] I have searched the existing issues and made sure this is not a duplicate.
- [X] I have provided sufficient information.

Issue #414, opened 2023-06-08
How can I upload captions/subtitles for videos?


chen-rn:
### Question

I understand that with the official youtube API v3 there is an endpoint that enables us to do that. I'm curious if that's achievable with YoutubeJs's current api set.

### Checklist

- [X] I am running the latest version.
- [X] I checked the documentation and found no answer.
- [X] I have searched the existing issues and made sure this is not a duplicate.
- [X] I have provided sufficient information.

chen-rn:
gotcha! I didn't understand the "Extending the library" section of the docs.

Is this something possible for me to simply extend the library?

chen-rn:
If this feature were to be implemented, would we be subject to the 10,000 API quote limit anyway? Or is there a way to bypass that.

PR #1231, opened 2026-08-12
feat(YouTube Studio Web): add support for YouTube Studio Web and primarily its uploading capabilities

### LINKED ISSUE #307: Enabling subtitles when uploading a video
### Describe your suggestion

I think it will be great if we can add the flags for enabling the subtitles for the video upload. 

The subtitles require the language to be set first, so a total of 2 things needs to be added.

Reference : 
image

image

### Other details

_No response_

### Checklist

### LINKED ISSUE #616: Can't upload videos larger than 2GB
### Steps to reproduce

Trying to upload big file (~8GB), got ERR_FS_FILE_TOO_LARGE error. Node.JS can't read whole file in memory.

### Failure Logs

```shell
RangeError [ERR_FS_FILE_TOO_LARGE]: File size (8453595338) is greater than 2 GiB
```

### Expected behavior

Should upload

### Current behavior

Getting error

### Version

Default

### Anything else?

_No response_

### Checklist

Illusion137:
A well-needed refactored port of my implementation of the YouTube Studio Web client, based off of [lib-origin](https://github.com/Illusion137/lib-origin/blob/dev/origin/src/youtube_studio/youtube_studio.ts).

**Noteable changes and features**
- YouTube Studio Web client
  - `uploadVideo()`
  - `updateVideo()`
  - `publishVideo()`
  - `uploadSubtitles()`
  - `uploadFeedback()`
  - ...some other small stuff
- Updated parser to include the full everything from the full sessionToken route
- Updated WEB_CREATOR version
- Added 2 example files to test out the code

**Testing**
I've added two example files that are easy enough to fill in:
`examples/youtube-studio-web/src/GetSessionToken.ts` - setups a BotGuard solver and grabs and logs your YouTube Session Token; important since this was really the only thing that was blocking using this client before.
`examples/youtube-studio-web/src/UploadVideo.ts` - same setup, but then uploads a video with a custom thumbnail and synced subtitles with 2 different progress callbacks.

**Final Notes**
I left a TODO in `src/types/BotGuard.ts` for: "make this more comprehensive" for the AttestationBinding. I think it might be better to have an interface like that, but currently it is kind of like a placeholder if the BotGuardSolver interface eventually expands.
Lastly, in `src/core/Actions.ts` there are some TODOs regarding the `session.context.request.eats`, which I'd really like some input of how it should be handled. As far I as know, `studio.youtube.com`:`ytcfg` provides an initial `EATS`, but it doesn't seem like it changes anything to have or not to have. Because of this I don't know if it's truly necessary to have an additional network request for nothing.

Fixes #307, #414, #418, #1164 
Likely fixes #616 

image
The uploading example after a successful run (redacted cookies and stuff)

<sub>Apologies for the large commits...</sub>


LuanRT:
Will convert this to a draft since there are likely some things we may want to address first.

src/parser/parser.ts
```diff
@@ -267,9 +274,15 @@ export function parseResponse<T extends IParsedResponse = IParsedResponse>(data:
     parsed_data.on_response_received_commands_memo = on_response_received_commands_memo;
   }
   _clearMemo();
+  if (Array.isArray(data.continuationContents)) {
+    parsed_data.upload_feedback_items = data.continuationContents
+      .map((entry) => entry.uploadFeedbackItemContinuation)
+      .filter((entry) => entry)
+      .map((entry) => new YTNodes.UploadFeedbackItem(entry));
```
LuanRT: If you need to create it manually like this (Which is by itself incorrect and inconsistent with the rest of the parser. Avoid using stuff from the `YTNodes` map internally), then it's not a real renderer/viewmodel.

The parsing logic so far seems a bit over the place. I can help fix and organize that when I get the time to test this branch.

edit:
also, the multiple iterations aren't good, memory usage wise

src/parser/parser.ts
```diff
@@ -267,9 +274,15 @@ export function parseResponse<T extends IParsedResponse = IParsedResponse>(data:
...
+  }
 
   _createMemo();
-  const continuation_contents = data.continuationContents ? parseLC(data.continuationContents) : null;
+  const continuation_contents = (data.continuationContents && !Array.isArray(data.continuationContents)) ? parseLC(data.continuationContents) : null;
```
LuanRT: What's up with the additional `isArray` check?

src/parser/parser.ts
```diff
@@ -530,6 +579,27 @@ export function parseResponse<T extends IParsedResponse = IParsedResponse>(data:
     parsed_data.target_id = data.targetId;
   }
 
+  if (data.videoId) {
+    parsed_data.video_id = data.videoId;
+  }
+
+  if (data.translation) {
+    parsed_data.translation = new YTNodes.Translation(data.translation);
```
LuanRT: .

src/parser/parser.ts
```diff
@@ -774,6 +844,27 @@ export function parseFormats(formats: RawNode[], this_response_nsig_cache: Map<s
   return formats?.map((format) => new Format(format, this_response_nsig_cache)) || [];
 }
 
+// useful for YTStudio since it outputs more in a nested data format rather than a 'rendering' format
+export function parseObject<T = any>(data: unknown): T {
```
LuanRT: Probably unneeded (does too much)

src/parser/parser.ts
```diff
@@ -530,6 +579,27 @@ export function parseResponse<T extends IParsedResponse = IParsedResponse>(data:
...
+  if (data.videoId) {
+    parsed_data.video_id = data.videoId;
+  }
+
+  if (data.translation) {
+    parsed_data.translation = new YTNodes.Translation(data.translation);
+  }
+
+  if (data?.creatorEntities?.wrappedVideoData?.video) {
+    parsed_data.creator_video = new YTNodes.CreatorVideo(data.creatorEntities.wrappedVideoData.video);
```
LuanRT: .

src/parser/parser.ts
Illusion137: the `/feedback` endpoint `continuationContents` looks a bit weird compared to others, especially since its a array. So the check is to use the code from above to set `parsed_data.upload_feedback_items` instead of `parsed_data.continuation_contents`. Well that was the intention, but I suppose yeah, it's unnecessary because of `parseLC` just returning null in the case. My bad.

src/parser/parser.ts
Illusion137: Will refactor this to use `parseArray` like the rest instead, the issue I was mainly thinking about not duplicating code since `/createvideo` has `uploadFeedbackItemRenderer` which is the same type as `/feedback`'s `uploadFeedbackItemContinuation`...

src/parser/parser.ts
```diff
@@ -232,8 +232,15 @@ export function getDynamicParsers() {
 export function parseResponse<T extends IParsedResponse = IParsedResponse>(data: IRawResponse): T {
   const parsed_data = {} as T;
 
+  const upload_feedback_item = (data.contents && !Array.isArray(data.contents) && data.contents.uploadFeedbackItemRenderer) ?
+    new YTNodes.UploadFeedbackItem(data.contents.uploadFeedbackItemRenderer) :
```
LuanRT: This is a problem too, since the parser already takes care of `contents` elsewhere.

src/parser/parser.ts
LuanRT: I think the biggest problem here is that `upload_feedback_items` and `upload_feedback_item` aren't really part of a real InnerTube response. Maybe `parseLC` should simply be updated to support array of continuations.

YouTube.js' parser is client agnostic. It shouldn't assume anything about any specific client, or care about the shape of the response.

(this is why it hardly ever breaks when yt changes a random prop somewhere, direct prop access is avoided outside of yt nodes and other parser classes).

Illusion137: > Maybe parseLC should simply be updated to support array of continuations

An approach like this does work I guess, but it does break code in a few spots that certain spots that access `continuation_contents`, moreover I feel like this would break other peoples codebases if something like this was added, so honestly I'm a bit stuck here...
```ts
export type ContinuationContents = null | ItemSectionContinuation | SectionListContinuation | LiveChatContinuation | 
  MusicPlaylistShelfContinuation | MusicShelfContinuation | GridContinuation | 
  PlaylistPanelContinuation | ContinuationCommand | UploadFeedbackItemContinuation;
export function parseLC(data: RawNode): ContinuationContents
export function parseLC(data: RawNode[]): ContinuationContents[]
export function parseLC(data: RawNode|RawNode[]): ContinuationContents|ContinuationContents[] {
  if (!Array.isArray(data)) {
    if (data.itemSectionContinuation)
      return new ItemSectionContinuation(data.itemSectionContinuation);
    if (data.sectionListContinuation)
      return new SectionListContinuation(data.sectionListContinuation);
    if (data.liveChatContinuation)
      return new LiveChatContinuation(data.liveChatContinuation);
    if (data.musicPlaylistShelfContinuation)
      return new MusicPlaylistShelfContinuation(data.musicPlaylistShelfContinuation);
    if (data.musicShelfContinuation)
      return new MusicShelfContinuation(data.musicShelfContinuation);
    if (data.gridContinuation)
      return new GridContinuation(data.gridContinuation);
    if (data.playlistPanelContinuation)
      return new PlaylistPanelContinuation(data.playlistPanelContinuation);
    if (data.continuationCommand)
      return new ContinuationCommand(data.continuationCommand);
    if (data.uploadFeedbackItemContinuation)
      return new YTNodes.UploadFeedbackItem(data.uploadFeedbackItemContinuation);
    return null;
  } 
  return data.map((node: RawNode) => parseLC(node));
}
```

edit:
I think actually it's probably better to just add to `parsed_data` a property called `continuation_contents_array` or something like that so something this small doesn't change many things?

src/parser/parser.ts
Illusion137: I think the other stuff is resolved now. For this I'll just go ahead and instead follow the example from `src/parser/classes/misc/Format.ts` and do all the parsing manually and also strip parts of the very long enums

PR #1239, opened 2026-08-13
refactor(VideoInfo): return a WatchNextContinuation instead of mutating state

### LINKED ISSUE #1237: (`VideoInfo.ts`) Stop mutating `watch_next_feed` when fetching continuations
### Describe your suggestion

This is some pretty crappy code we wrote years ago, and it doesn't play nice with frameworks like Vue for example. It would be better to create a separate class for it, similar to how comment thread replies are handled:
https://github.com/LuanRT/YouTube.js/blob/main/src/parser/classes/comments/CommentThread.ts#L99-L117
https://github.com/LuanRT/YouTube.js/blob/main/src/parser/classes/misc/CommentsContinuation.ts#L1-L61

### Other details

_No response_

### Checklist

- [x] I am running the latest version.
- [x] I checked the documentation and found no answer.
- [x] I have searched the existing issues and made sure this is not a duplicate.
- [x] I have provided sufficient information.

MFA-G:
Fixes #1237

## What

`VideoInfo#getWatchNextContinuation()` mutated the instance in place: it overwrote `watch_next_feed`, reassigned the private `#watch_next_continuation`, and returned `this`. As noted in the issue, that pattern doesn't play well with reactive frameworks (Vue and friends) because the object identity stays the same while its contents change underneath.

This adds a `WatchNextContinuation` class modelled directly on the existing `CommentsContinuation` / `CommentThread#getContinuation()` pattern referenced in the issue:

```ts
export default class WatchNextContinuation {
  public contents: ObservedArray<YTNode>;

  get has_continuation(): boolean;
  public async getContinuation(): Promise<WatchNextContinuation>;
}
```

`VideoInfo#getWatchNextContinuation()` now just builds one from the response and returns it; `VideoInfo` itself is left untouched.

Two small behavioural notes:

* The `ContinuationItem` is **filtered out** of `contents` rather than `pop()`ed off after the fact, so feed items are never mixed with the continuation marker (the old code also relied on it being the last element).
* An `InnertubeError` is thrown when `AppendContinuationItemsAction` is missing, same as before.

I deliberately left `VideoInfo`'s constructor, `watch_next_feed`, and `wn_has_continuation` alone — the initial feed still comes from the watch page as it always did, so nothing changes for the first page.

## Breaking change

`getWatchNextContinuation()` used to resolve to the (mutated) `VideoInfo`. It now resolves to a `WatchNextContinuation`:

```ts
// before
const info = await yt.getInfo(id);
let feed = info.watch_next_feed;
while (info.wn_has_continuation) {
  await info.getWatchNextContinuation();
  feed = info.watch_next_feed; // same object, mutated
}

// after
const info = await yt.getInfo(id);
let page = await info.getWatchNextContinuation();
console.log(page.contents);
while (page.has_continuation) {
  page = await page.getContinuation();
  console.log(page.contents);
}
```

Note that `YTShorts.ShortFormVideoInfo#getWatchNextContinuation()` has the same mutation pattern but a different response shape (`reel_watch_sequence` / `ContinuationCommand`), so I left it out of this PR to keep the change focused. Happy to follow up on it if you'd like it aligned too.

## Testing

* `npx tsc --noEmit -p tsconfig.json` — clean
* `npx eslint src/` — clean
* `npm run build:esm` and `npm run build:parser-map` — clean (the new class is exported from `src/parser/misc.ts` by the generator, as with `CommentsContinuation`)
* Ran the built class against a hand-crafted `appendContinuationItemsAction` payload to verify behaviour end to end:

```
contents length: 2
types: CompactVideo, CompactVideo
has_continuation: true
no ContinuationItem leaked: true
last page has_continuation: false len: 2
throws on missing action: InnertubeError
```

The `tests/main.test.ts` suite hits the live API and the only `getWatchNextContinuation` case there is the Shorts one, which is unaffected.

### DIFF
--- a/src/parser/classes/misc/WatchNextContinuation.ts
+++ b/src/parser/classes/misc/WatchNextContinuation.ts
@@ -1,0 +1,54 @@
+import { InnertubeError } from '../../../utils/Utils.js';
+import { type ObservedArray, observe, type YTNode } from '../../helpers.js';
+import AppendContinuationItemsAction from '../actions/AppendContinuationItemsAction.js';
+import ContinuationItem from '../ContinuationItem.js';
+
+import type { Actions } from '../../../core/index.js';
+import type { INextResponse } from '../../types/index.js';
+
+export default class WatchNextContinuation {
+  public contents: ObservedArray<YTNode>;
+
+  readonly #actions: Actions;
+  readonly #continuation?: ContinuationItem;
+
+  constructor(actions: Actions, data: INextResponse) {
+    this.#actions = actions;
+
+    const append_continuation_items_action = data.on_response_received_endpoints?.firstOfType(AppendContinuationItemsAction);
+
+    if (!append_continuation_items_action)
+      throw new InnertubeError('AppendContinuationItemsAction not found', data);
+
+    const items: YTNode[] = [];
+
+    for (const item of append_continuation_items_action.contents) {
+      if (item.is(ContinuationItem)) {
+        this.#continuation = item;
+      } else {
+        items.push(item);
+      }
+    }
+
+    this.contents = observe(items);
+  }
+
+  /**
+   * Indicates whether there are more items that can be fetched.
+   */
+  get has_continuation(): boolean {
+    return !!this.#continuation;
+  }
+
+  /**
+   * Retrieves the next batch of items.
+   */
+  public async getContinuation(): Promise<WatchNextContinuation> {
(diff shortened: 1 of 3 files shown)


MFA-G:
Still relevant — the branch is mergeable against current `main` and CI is green. Happy to rebase or adjust the approach if you'd prefer a different shape for the continuation API.

